Issue #12, opened 2014-09-10
[RFC] Improve support for auto-unsubscribing observables and newer language/API levels


mttkay:
Currently, using Observables in Android components that have access to `Context` (or are themselves) requires one to carefully think about detaching from these sequences as part of the component life-cycle, since otherwise a closure the sequence holds a (strong) reference to might leak the current context until the time the sequence finishes and releases all subscribers.

A good example are configuration changes, where Android first destroys, then recreates an Activity, but observables might still be holding on to a subscriber created within the scope of that Activity. Even with retained fragments that detach from their host activity, it's easy enough to leak the attached context indirectly by holding on to a view first created through that context (every view in Android has a strong reference to the Activity context it was first created in.)

This ticket aims to find solutions to improve this situation, potentially by leveraging newer Android API levels where available, and/or add first class support for newer Java language levels (Java 8/Retrolambda)

Some suggestions that have been made already follow.

**Android 14+ APIs**

One option would be to make use of Activity life cycle callbacks:
http://developer.android.com/reference/android/app/Application.ActivityLifecycleCallbacks.html

This might allows us to unsubscribe once Android is about to destroy an activity.

**Java 8 lambdas and Retrolambda**

Part of the problem is that every "lambda" in Java 7 or below is an anonymous inner class that will hold a strong reference to the outer class, even if it's a pure function that does not access state outside the lambda. Java 8 + Retrolambda could help here, since pure functions will be materialized as static methods that hold no reference back to their owner.

http://cr.openjdk.java.net/~briangoetz/lambda/lambda-translation.html
https://github.com/orfjackal/retrolambda

**Weak subscribers**

We've taken several stabs at this and always dismissed the solution. The idea was to use WeakReferences to bind subscribers and lambdas. The problem with his approach is that it doesn't work with lambdas, since the only incoming reference would be weak, and they are eligible for GC the instant they're created. There might still be other ways to achieve this, however, I'll leave it open for discussion.

hamidp:
Do we consider https://github.com/ReactiveX/RxAndroid/pull/75 as good enough of a solution for now?

JakeWharton:
[...]

Issue #3, opened 2014-08-23
Address the problem of asynchronously unsubscribing in life-cycle callbacks


mttkay:
This was first reported here: https://github.com/ReactiveX/RxJava/issues/1590

Opening a new issue here since RxAndroid is now separated out of core.

Potential solutions so far:

1. Always use `bindFragment` when subscribing Subscribers in fragments that access views in `onNext` (defensive programming style)
1. Provides a safe UI subscriber class that only propagates notifications if not `isUnsubscribed`
1. ... ?

PR #13, opened 2014-09-14
[review-only] OperatorSynchronousUnsubscribe

nsk-mironov:
Possible solution for the problem discussed here https://github.com/ReactiveX/RxAndroid/issues/3 and here https://github.com/ReactiveX/RxAndroid/issues/12.

`OperatorSynchronousUnsubscribe` ensures that a `Subscriber` passed to `Observable.subscribe` won't receive any notifications after `Subscription.unsubscribe` is called. It also releases a reference to prevent memory leaks.

In or order to work properly `OperatorSynchronousUnsubscribe` **must** be the latest lifted operator:

``` java
final Subscription subscription = ViewObservable.clicks(button, false)
        .observeOn(AndroidSchedulers.mainThread())
        .lift(new OperatorSynchronousUnsubscribe<View>())
        .subscribe(new Action1<View>() {
            @Override
            public void call(View view) {
                Log.d("RxAndroid", "Button Clicked");
            }
        });
```

/cc @mttkay

### DIFF
--- a/src/main/java/rx/operators/OperatorSynchronousUnsubscribe.java
+++ b/src/main/java/rx/operators/OperatorSynchronousUnsubscribe.java
@@ -1,0 +1,63 @@
+package rx.operators;
+
+import rx.Observable;
+import rx.Observer;
+import rx.Subscriber;
+import rx.functions.Action0;
+import rx.observers.EmptyObserver;
+import rx.subscriptions.Subscriptions;
+
+public class OperatorSynchronousUnsubscribe<T> implements Observable.Operator<T, T> {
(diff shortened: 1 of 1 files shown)


johnwatsondev:
@nsk-mironov  Hi , is there a better elegant way in the lastest RxJava ? I have use your approach and it worked. Thanks for sharing !

pyricau:
This is great!

AllenVork:
@nsk-mironov  I'm using :
    compile 'io.reactivex:rxandroid:1.2.1'
    compile 'io.reactivex:rxjava:1.3.0'
    Why can't I find the OperatorSynchronousUnsubscribe ?

Issue #59, opened 2014-11-19
Reorganizing the samples module


mttkay:
Now that the samples module is back, it's clear this needs working on. There should be structure to the samples app instead of being a bunch of unrelated classes. Any suggestions for how this should look like?

For instance:
- have a landing activity that let's you navigate samples a la Android SDK samples (can keep it simple for now)
- organize samples by functional group (view bindings such as text observables, scroll observables etc., basic functionality such as life-cycle binding, working with cache/replay, ...)

I would also like to propose that every PR introducing new functionality to the library or a sub module should include a sample demonstrating use and doubling as documentation of the added feature.

JakeWharton:
As part of #172 the sample was gutted. As of now, this isn't needed, and the rise of more comprehensive Rx-ified examples in the last few months also helps here.

PR #82, opened 2014-11-25
Re organize sample

zsiegel:
Starting work on #59


JakeWharton:
As part of #172 the sample has been gutted and no longer needs reorganizing. Thanks for your efforts though.

PR #84, opened 2014-11-25
Adding a ReactiveNavigator to handle toActivityForResult in an Observable way

Dorvaryn:
The aim of this component is to handle the navigation to an Activity for result using an Observable and thus avoiding the need to check for result/response code and all the boilerplate code related.

### DIFF
--- a/rxandroid-framework/build.gradle
+++ b/rxandroid-framework/build.gradle
@@ -1,6 +1,17 @@
 // Multi-module modules aren't really supported
 version = rootProject.version
 
+configurations {
+    // This is a rough approximation of Android plug-in's 'provided' configuration, where
+    // you can give compile-time only dependencies.
+    compileOnly
+}
+
+dependencies {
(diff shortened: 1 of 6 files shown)


Dorvaryn:
@hamidp Yes, sorry about the delay I had some family issues and had to tune off from work and OSS stuff for a while.

hamidp:
:+1: Provided we can get some docs on it as well.

Dorvaryn:
@hamidp What type of docs are you thinking of ? I tried to JDoc the code to make it simple enough to understand. There is one more PR adding the rotation handling code and then I have a sample project that presents how to use those tools. I was thinking of adding a doc for usage at that point.

hamidp:
@Dorvaryn More along the lines of a sample readme. 

My :+1: isn't contingent on that -- just would be nice to let consumers of the library know we've got these awesome things.

Dorvaryn:
Should we get this merged in ?

PR #100, opened 2014-11-28
Reverted some View and Widget Observables to use simple objects instead of Event objects

Alexander--:
Recent changes to UI observables (particularly, 406bf6840fff2a98316faec8702311f9322047aa) added number of POJO event classes as well as some of associated infrastructure. Some of those were really useful, but few others: click, text change and toggle (aka "input") event classes look forced, as if those were added solely to match style of similar methods. Unfortunately, as explained below, they aren't very useful and accomplish nothing, except adding redundant entities and increasing range of possible issues.

Few possible uses for event classes, that come to mind are:

1) Encapsulating state, not stored in event source itself. This is the only sensible use for event classes in Android, which is why OnItemClickEven and OnListViewScrollEvent are legitimately useful and were added for good reason. Clearly not the case for click, text change and CompoundButton toggle events.
2) Storing associated data, for example for doing `reduce` on it. None of the event classes in question (except OnSubscribeTextViewInput) contain any useful state. Click and compound button switch are self-descriptive. CharSequence, returned by `TextView.getText()`, can be mutable, and as such, is dangerous to use, unless immediately converted to string (which means, that at least one `map` call have to be done anyway).
2) Turning emission of same object into emission of multiple distinct objects. RxJava already have `timestamp` operator for this.
3) Passing those events somewhere else, possibly after serializing them. Doing so with all 3 events in question is pointless and even potentially dangerous. All of them contain little besides strong reference to event source, which makes them potential source of memory leaks, and makes it harder to write Rx plugins, detecting those.

This merge request adds back Observables with tests and corresponding static methods, emitting plain View, TextView and Boolean objects. It also deprecates event-based methods. While we are at it, I have also renamed "input" to "toggle" as original naming choice was just plain bad.

### DIFF
--- a/rxandroid/src/main/java/rx/android/view/OnCheckedChangeEvent.java
+++ b/rxandroid/src/main/java/rx/android/view/OnCheckedChangeEvent.java
@@ -16,6 +16,8 @@ package rx.android.view;
 import android.widget.CompoundButton;
 import com.google.auto.value.AutoValue;
 
+/** @deprecated this class will be removed soon */
+@Deprecated
 @AutoValue
 public abstract class OnCheckedChangeEvent {
     public abstract CompoundButton view();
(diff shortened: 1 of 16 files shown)


JakeWharton:
Per #172 we have removed the view binding observables for the next release.

PR #103, opened 2014-11-29
First attempt at adding Account Observable

Alexander--:
This is a rough draft, produced from code, I have been using for a while already.

Note, that unlike most other operators in the library `OnSubscribeAddAccount` is left public to allow people to better adjust it for various needs - the code is already too sophisticated and does too much to try and stuff everything in bunch of static methods.

Before finishing this, I would like to know answers to following questions:

1) How do I test it? Robolectric 2.3 does not have proper shadow for AccountManager, so should I just write my own? By the way, are there plans to update to Robolectric 2.4, perhaps by employing [SDK manager plugin](https://github.com/JakeWharton/sdk-manager-plugin) to fetch dependencies? I am aware of [#1369](https://github.com/robolectric/robolectric/issues/1369), but at current rate it is not going resolve itself.
2) What are project guidelines (ok, what _should_ those be) regarding use of annotations? Complex AccountManager methods need `@Nullable` annotation rather badly, but which one to choose? Android ones? IntelliJ ones (used so far, because they are easiest to get). Perhaps, something more complex (Checker Framework)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,6 @@ jdk:
 android:
   components:
     - android-20
+    - extra-android-m2repository
 
 script: "./gradlew build"
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,6 @@ buildscript {
     // The above dependency will bring in guava, which then pollutes child projects
     force 'com.google.guava:guava:15.0'
   }
-
 }
 
 apply plugin: 'rxjava-project'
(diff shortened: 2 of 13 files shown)


JakeWharton:
As part of #172 support for components like this have been removed. There's far too much breadth in the use case of these components to warrant inclusion in a core module like RxAndroid. This is something that would be better served as a standalone, focused library.

Issue #117, opened 2014-12-29
Release 0.24.0


hamidp:
Seems like a good time to do a release and get some more movement on the pending PRs.

Thoughts?

iNoles:
AppObservable is much easier for any android developers without depend on lifecycle

JakeWharton:
@mttkay got hit by a bus so we can't release. We have a very low bus factor on this project.

mttkay:
I can't release either, since it goes through Netflix's build infra. You
will have to talk to Ben. He did all the releases for us so far.

JakeWharton:
Will do. Thanks :+1:

benjchristensen:
Doing so now.

benjchristensen:
The Gradle/CloudBees config has something wrong so need to figure that out. /cc @quidryan Seems we need to upgrade to v2.x of the nebula plugin.

benjchristensen:
@quidryan I tried upgrading to the new Nebula plugins and Gradle 2:

```
    classpath 'com.netflix.nebula:gradle-rxjava-project-plugin:2.+'
    classpath 'com.netflix.nebula:gradle-extra-configurations-plugin:2.+'
```

This then results in this failure:

```
* Where:
Build file '/Users/benjchristensen/development/github/ReactiveX/RxAndroid/sample-app/build.gradle' line: 12

* What went wrong:
A problem occurred evaluating project ':sample-app'.
> Failed to apply plugin [id 'com.android.application']
   > Gradle version 1.10 is required. Current version is 2.2.1. If using the gradle wrapper, try editing the distributionUrl in /Users/benjchristensen/development/github/ReactiveX/RxAndroid/gradle/wrapper/gradle-wrapper.properties to gradle-1.10-all.zip
```

But I believe I need to be upgraded to `gradle-rxjava-project-plugin:2.+` in order for the new CloudBees config to work, and that requires Gradle 2.

benjchristensen:
Is it expected to have a new artifact `rxandroid-framework` released to Maven Central?

hamidp:
Yes.

benjchristensen:
Thanks for the fast response @hamidp

benjchristensen:
0.24.0 has been released. It is on its way to Maven Central and can be seen on BinTray at https://bintray.com/reactivex/RxJava/RxAndroid/view

I had to disable the `sample-app` module to get through the build process as I had to upgrade to Gradle 2. We can now deal with that issue separately.

[...]

PR #118, opened 2015-01-03
Upgrade Nebula Plugin and Gradle

benjchristensen:
The CloudBees build/release config has been upgraded to assuming Nebula RxJava plugin 2.x which assumes Gradle 2. This is an attempt to upgrade this project. 

See https://github.com/ReactiveX/RxAndroid/issues/117


settings.gradle
```diff
@@ -2,4 +2,4 @@ rootProject.name='rxandroid-root'
 
 include 'rxandroid'
 include 'rxandroid-framework'
-include 'sample-app'
+//include 'sample-app'
```
dpsm: Intentional?

benjchristensen: Yes, intentional to try and get 0.24 released until we can figure out how to have the `sample-app` working with the latest build plugin.

PR #121, opened 2015-01-10
Fixing android gradle version ref so that sample apps can be built

greg7gkb:
@benjchristensen: fixed

### DIFF
--- a/sample-app/build.gradle
+++ b/sample-app/build.gradle
@@ -1,7 +1,8 @@
 buildscript {
     repositories { jcenter() }
     dependencies { 
-        classpath 'com.android.tools.build:gradle:0.12.+' // 0.12.x is for Gradle 1.12 projects
+        classpath 'com.android.tools.build:gradle:1.+' // Android gradle 1.+ is for Gradle 2.+ projects
+        projects
     }
 }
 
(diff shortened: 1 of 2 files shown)


sample-app/build.gradle
```diff
@@ -1,7 +1,8 @@
 buildscript {
     repositories { jcenter() }
     dependencies { 
-        classpath 'com.android.tools.build:gradle:0.12.+' // 0.12.x is for Gradle 1.12 projects
+        classpath 'com.android.tools.build:gradle:1.+' // Android gradle 1.+ is for Gradle 2.+ projects
```
JakeWharton: Remove the comment its redundant. Also please specify the while version. Wildcard dependencies are an atrocity

greg7gkb: Comment is not redundant because otherwise it's unclear how then android gradle version maps to standard gradle instals.
Understand the wildcard dep, fixing...

JakeWharton: OK well the comment is not accurate. Version 1.0.0 cannot be used with
Gradle 2.0 or 2.1 and it most certainly will be compatible with 3.x.
Previous versions worked across the 1.x and 2.x divide.

greg7gkb: Okay, agree with that.  Stripping it.

greg7gkb:
I put the runProguard removal in the wrong branch, and this is what's causing CI to fail.
Merging both pull requests should fix the issue.

PR #122, opened 2015-01-10
Updating sample app sdk and build-tools version numbers for Lollipop

greg7gkb:


sample-app/build.gradle
```diff
@@ -12,18 +12,18 @@ repositories {
...
+        targetSdkVersion 21
         versionCode 1
         versionName "1.0"
     }
     buildTypes {
         release {
-            runProguard false
+            minifyEnabled false // Replaces 'runProguard false'
         }
     }
```
JakeWharton: You can delete this whole block. False is the default.

PR #123, opened 2015-01-10
Updating android gradle version ref

greg7gkb:
Fixed previous pull req


JakeWharton:
So glad you did this. Thanks!

greg7gkb:
Np.  I was just trying to get the project setup so I could put some real work in :)
Will start with the current set of issues and let you know if I think of new things to add.

Issue #95, opened 2014-11-27
"bindFragment" Method Signature


JakeWharton:
It takes `Object` and uses ugly `instanceof` and runtime API checks. Why not just have two overloads?

JakeWharton:
Ok, the way it stands now this will actually fail at compilation time, not runtime (see #136).

This is because the methods become "pure" overloads of each other. That is, they have the same name, return type, number of parameters, and type of second parameter. They only differ by the type of the first parameter.

In the very few times I've done this, the overloads have differed in some way even if that difference produces the same API at the call site. A prime example: varargs.

Changing the support-v4 variant's method signature to the following:
```java
bindFragment(android.support.v4.app.Fragment fragment, Observable<T> source, Object... dummy)
```
actually works! You can also rename the method to `bindSupportFragment` and it will work. Neither approach fails at runtime.

So, which do we like?

rosshambrick:
A little late for a reply I guess, but I'd prefer `bindFragment(...)` for two reasons.  1- It wouldn't be a breaking change and 2 - the addition of 'Support' doesn't add much value.  This is assuming it technically works just as well either way.

JakeWharton:
Per my comment above, it does not work.

rosshambrick:
Ah yes, I missed `Object... dummy` the first time somehow. Carry on

PR #136, opened 2015-02-25
Don't throw exception, instead user should find out at compile time that they're passing the wrong argument

premnirmal:
Inside AppObservable, you have a method `bindFragment(Object,Observable<T>)`. Instead of accepting an `Object` parameter, you should have two methods, one accepting a `Fragment` and the other accepting a `android.support.v4.app.Fragment`.

### DIFF
--- a/rxandroid/src/main/java/rx/android/app/AppObservable.java
+++ b/rxandroid/src/main/java/rx/android/app/AppObservable.java
@@ -110,9 +110,7 @@ public final class AppObservable {
      * @param source   the source sequence
      */
     public static <T> Observable<T> bindFragment(Fragment fragment, Observable<T> source) {
-        Assertions.assertUiThread();
-        final Observable<T> o = source.observeOn(mainThread());
-        return o.lift(new OperatorConditionalBinding<T, Fragment>(fragment, FRAGMENT_VALIDATOR));
+        return bindFragmentInternal(fragment, source);
     }
 
     /**
@@ -130,8 +128,6 @@ public final class AppObservable {
      * @param source   the source sequence
      */
     public static <T> Observable<T> bindFragment(android.support.v4.app.Fragment fragment, Observable<T> source) {
-        Assertions.assertUiThread();
-        final Observable<T> o = source.observeOn(mainThread());
-        return o.lift(new OperatorConditionalBinding<T, android.support.v4.app.Fragment>(fragment, FRAGMENTV4_VALIDATOR));
+        return bindFragmentInternal(fragment, source);
     }
 }
--- a/rxandroid/src/main/java/rx/android/app/AppObservable.java
+++ b/rxandroid/src/main/java/rx/android/app/AppObservable.java
@@ -15,7 +15,6 @@ package rx.android.app;
 
 import android.app.Activity;
 import android.app.Fragment;
-import android.os.Build;
 
 import rx.Observable;
 import rx.android.internal.Assertions;
@@ -62,10 +61,10 @@ public final class AppObservable {
 
     /**
      * Binds the given source sequence to an activity.
-     * <p>
+     * <p/>
      * This helper will schedule the given sequence to be observed on the main UI thread and ensure
      * that no notifications will be forwarded to the activity in case it is scheduled to finish.
-     * <p>
+     * <p/>
      * You should unsubscribe from the returned Observable in onDestroy at the latest, in order to not
      * leak the activity or an inner subscriber. Conversely, when the source sequence can outlive the activity,
      * make sure to bind to new instances of the activity again, e.g. after going through configuration changes.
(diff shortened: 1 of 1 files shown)


mttkay:
I've had the same discussion with @JakeWharton before. I'm still fairly sure I did this because if the support library is not on the classpath, this will crash at runtime when the class is loaded. Jake though otherwise. Not sure if you verified that this will actually work when you do not link the support library?

premnirmal:
@mttkay Isn't the support library already int he classpath https://github.com/ReactiveX/RxAndroid/blob/0.x/build.gradle#L26 ?

rxandroid/src/main/java/rx/android/app/AppObservable.java
```diff
@@ -62,10 +61,10 @@ public Boolean call(android.support.v4.app.Fragment fragment) {
 
     /**
      * Binds the given source sequence to an activity.
-     * <p>
+     * <p/>
```
JakeWharton: This is wrong, revert. Paragraphs are not self-closing and doclint on JDK 8 will fail the build for this.

rxandroid/src/main/java/rx/android/app/AppObservable.java
```diff
@@ -79,13 +78,30 @@ public Boolean call(android.support.v4.app.Fragment fragment) {
         return source.observeOn(mainThread()).lift(new OperatorConditionalBinding<T, Activity>(activity, ACTIVITY_VALIDATOR));
     }
 
+    @Deprecated
+    public static <T> Observable<T> bindFragment(Object fragment, Observable<T> source) {
+        return bindFragmentInternal(fragment, source);
+    }
+
+    private static <T> Observable<T> bindFragmentInternal(Object fragment, Observable<T> source) {
```
JakeWharton: Destroy this method and move the respective implementations to the appropriate methods.

rxandroid/src/main/java/rx/android/app/AppObservable.java
```diff
@@ -79,13 +78,30 @@ public Boolean call(android.support.v4.app.Fragment fragment) {
         return source.observeOn(mainThread()).lift(new OperatorConditionalBinding<T, Activity>(activity, ACTIVITY_VALIDATOR));
     }
 
+    @Deprecated
+    public static <T> Observable<T> bindFragment(Object fragment, Observable<T> source) {
```
JakeWharton: Just delete. The overloads will resolve on upgrade for anyone using this correctly. And if you weren't then we just turned your runtime failure to a compile time failure.

JakeWharton:
This will crash on Android 2.1 or earlier only.

JakeWharton:
Easy to verify: just create a new empty project without the support library and call `bindFragment(android.app.Fragment)` and run it on any new OS.

mttkay:
>  Isn't the support library already in the classpath https://github.com/ReactiveX/RxAndroid/blob/0.x/build.gradle#L26 ?

It's linked with `provided` scope. That means, it's in the classpath visible to the respective compilation tasks in Gradle, but it will not be available at runtime for any clients depending on the RxJava library. The way I read this is: if the class loader loads this class, and finds a reference to a symbol from the (missing) support library, it will shake its fist at you and crash.

If, however, the method signature declares `Object`, and we do the check/cast lazily upon method invocation, we can prevent this from happening since the code path will never execute unless the support Fragments are available. Does that make sense?

JakeWharton:
> if the class loader loads this class, and finds a reference to a symbol from the (missing) support library, it will shake its fist at you and crash.

This is the behavior that changed with Android 2.2 and is no longer true.

mttkay:
That's not to say that I'd love to see code like this go. I guess I just couldn't come up with a better solution when I first wrote this, so by all means please rewrite it. Just do it in a way that doesn't force support-v* on people.

mttkay:
> This is the behavior that changed with Android 2.2 and is no longer true.

Right, you mentioned that before. Fair enough then.

rxandroid/src/main/java/rx/android/app/AppObservable.java
premnirmal: done, thanks

premnirmal:
> if the class loader loads this class, and finds a reference to a symbol from the (missing) support library, it will shake its fist at you and crash.

> > This is the behavior that changed with Android 2.2 and is no longer true.

@mttkay @JakeWharton , yes you're right. It won't crash, but it won't compile because of the method signature has the support fragment, which isn't included in the classpath.

> Easy to verify: just create a new empty project without the support library and call bindFragment(android.app.Fragment) and run it on any new OS.
Tried it, won't compile because of the `bindFragment(android.support.v4.app.Fragment)` method.

Thanks for confirming.

Closing now.

JakeWharton:
It's already compiled in the class file. The problem (or rather, lack of one) would be in the source that was trying compile against this class file and link against its methods.

premnirmal:
yeah, the sample app is what doesn't compile

JakeWharton:
With what error?

premnirmal:
`Error:(77, 41) error: cannot access Fragment`
`class file for android.support.v4.app.Fragment not found`

on the line where I call `AppObservable.bindFragment(Fragment)`

premnirmal:
here https://github.com/ReactiveX/RxAndroid/blob/0.x/sample-app/src/main/java/rx/android/samples/ListeningFragmentActivity.java#L77

JakeWharton:
Interesting. Will try. We definitely do this exact thing without problems elsewhere.

JakeWharton:
Forcing conversation back to #95.

PR #140, opened 2015-03-02
add bindLifecycle rxactivity/rxfragment instance method

Yarikx:
`bindLifecycle(sourceObservable)` is just a alias for corresponding static method `LifecycleObservable.bindFragmentLifecycle(source)`/`LifecycleObservable.bindActivityLifecycle(source)`

Main purpose is to decrease boilerplate code while binding observable to lifecyle
Also, if method references are supported, this method can be used directly in `compose` operator: 
```java
source.compose(this::bindLifecycle)
```

Another issue it can prevent is copy/paste of code between fragment/activity which include `bindLifecycle`:

```java
LifecycleObservable.bindFragmentLifecycle(lifecycle(), source)
```
if this code was copied while refactoring to activity, it will compile but can lead to wrong behavior.
While:
```java
bindLifecycle(source)
```
will work as expected as proper bind strategy is specified in rxactivity/rxfragment

### DIFF
--- a/rxandroid-framework/src/main/java/rx/android/app/RxActivity.java
+++ b/rxandroid-framework/src/main/java/rx/android/app/RxActivity.java
@@ -17,6 +17,7 @@ package rx.android.app;
 import android.os.Bundle;
 import rx.Observable;
 import rx.android.lifecycle.LifecycleEvent;
+import rx.android.lifecycle.LifecycleObservable;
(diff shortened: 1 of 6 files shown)


JakeWharton:
As part of #172 these binding methods have been removed for a future release. You should unsubscribe explicitly in the lifecycle callback of your choice.

PR #145, opened 2015-03-10
Add helper contentObservable for contentObserver

HTChang:
Add an helper function 
```
ContentObservable.fromContentObserver(contentResolver, watchUri, schedulerHandler)
```
that makes the ContentObserver as an Observable

### DIFF
--- a/rxandroid/src/main/java/rx/android/content/ContentObservable.java
+++ b/rxandroid/src/main/java/rx/android/content/ContentObservable.java
@@ -13,11 +13,13 @@
  */
 package rx.android.content;
 
+import android.content.ContentResolver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.SharedPreferences;
 import android.database.Cursor;
+import android.net.Uri;
 import android.os.Handler;
 
 import rx.Observable;
@@ -25,6 +27,35 @@ import rx.Observable;
 public final class ContentObservable {
     private ContentObservable() {
         throw new AssertionError("No instances");
+    }
+
+    /**
+     * Create Observable that wraps ContentObserver and emits received changed Uris.
+     *
+     * @param contentResolver
+     * @param uri             The URI to watch for changes.  This can be a specific row URI,
+     *                        or a base URI for a whole class of content.
+     * @return
+     */
+    public static Observable<Uri> fromContentObserver(ContentResolver contentResolver, Uri uri) {
+        return Observable.create(new OnSubscribeContentObserverRegister(contentResolver, uri,
+            null));
+    }
+
+    /**
+     * Create Observable that wraps ContentObserver and emits received changed Uris.
+     *
+     * @param contentResolver
+     * @param uri              The URI to watch for changes.  This can be a specific row
+     *                         URI, or a base URI for a whole class of content.
+     * @param schedulerHandler The handler to run {@link android.database
+     *                         .ContentObserver#onChange(boolean)} on, or null if none.
+     * @return
+     */
+    public static Observable<Uri> fromContentObserver(ContentResolver contentResolver, Uri uri,
+        Handler schedulerHandler) {
+        return Observable.create(new OnSubscribeContentObserverRegister(contentResolver, uri,
+            schedulerHandler));
     }
 
     /**
--- a/rxandroid/src/main/java/rx/android/content/OnSubscribeContentObserverRegister.java
+++ b/rxandroid/src/main/java/rx/android/content/OnSubscribeContentObserverRegister.java
@@ -1,0 +1,64 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


rxandroid/src/main/java/rx/android/content/ContentObservable.java
```diff
@@ -13,21 +13,45 @@
...
+     * @param contentResolver
+     * @param uri The URI to watch for changes. This can be a specific row URI, or a base URI
+     * for a whole class of content.
+     * @param schedulerHandler Handler identifying the thread that will observe
+     *      the content changes.
+     */
+    static Observable<Uri> fromContentObserver(final ContentResolver contentResolver,
+        final Uri uri, final Handler schedulerHandler) {
+        if (SDK_INT < 16) {
+            return Observable.error(new UnsupportedOperationException());
```
JakeWharton: Just throw

HTChang: Thanks for reminder.

JakeWharton:
This would need a test. Not sure how to do it with Robolectric, but with an integration test you can extend `ProviderTestCase2`.

rxandroid/src/main/java/rx/android/content/ContentObservable.java
```diff
@@ -13,21 +13,45 @@
...
+     *
+     * @param contentResolver
+     * @param uri The URI to watch for changes. This can be a specific row URI, or a base URI
+     * for a whole class of content.
+     * @param schedulerHandler Handler identifying the thread that will observe
+     *      the content changes.
+     */
+    static Observable<Uri> fromContentObserver(final ContentResolver contentResolver,
+        final Uri uri, final Handler schedulerHandler) {
+        if (SDK_INT < 16) {
```
ronshapiro: Can you use `Build.VERSION_CODES.JELLY_BEAN` instead of `16`?

JakeWharton: Saw this too, but doesn't this project build with 4.0 because it's not an Android module?

rxandroid/src/main/java/rx/android/content/OnSubscribeContentObserverRegister.java
```diff
@@ -0,0 +1,64 @@
...
+ *  has been called.
+ */
+class OnSubscribeContentObserverRegister implements Observable.OnSubscribe<Uri> {
+
+    private final ContentResolver contentResolver;
+    private final Handler schedulerHandler;
+    private final Uri uri;
+
+    public OnSubscribeContentObserverRegister(ContentResolver contentResolver,
+        Handler schedulerHandler, Uri uri) {
```
JakeWharton: Needs 4 more spaces indented

rxandroid/src/main/java/rx/android/content/OnSubscribeContentObserverRegister.java
```diff
@@ -0,0 +1,64 @@
...
+import rx.Observable;
+import rx.Subscriber;
+import rx.Subscription;
+import rx.functions.Action0;
+import rx.subscriptions.Subscriptions;
+
+/**
+ *  Register a {@link ContentObserver} and make it as a ContentObservable that
+ *  emits {@link Uri} when {@link ContentObserver#onChange(boolean, android.net.Uri)}
+ *  has been called.
```
JakeWharton: Eliminate the double-spaces before each line here.

rxandroid/src/main/java/rx/android/content/ContentObservable.java
ronshapiro: Meh, you're right. That's annoying. I guess this could be reversed to `SDK_INT > VERSION_CODES.ICE_CREAM_SANDWICH_MR1`, not sure if that feels weird though.

JakeWharton: There's a 4.1.1.4 in Maven Central that can be used to get `JELLY_BEAN`

ronshapiro:
[ShadowContentObserver#dispatchChange](http://robolectric.org/javadoc/org/robolectric/shadows/ShadowContentObserver.html#dispatchChange(boolean)) might be a good place to look for tests.

JakeWharton:
Yeah, but Robolectric :(

wuman:
An improved version here https://github.com/wuman/RxAndroid/commit/c85d2ffa0058039a0a16f2472bbd95af5b701403.  Didn't want to send a separate pull request because it's largely based on this PR.

Improvements include:
- Removed the need to check for api levels.  For api levels below 16, the base Uri will be emitted.
- Added a unit test.

@HTChang Feel free to incorporate my changes as you see fit.

HTChang:
@wuman  thanks, already modified my code per your suggestion!

artem-zinnatullin:
Damn, I just needed this too and implemented it when I saw this PR..

But, I have an addition:

`ContentResolver#registerContentObserver()` has parameter `notifyForDescendents`, which is just true in this PR, I think it should be decision of the user

From Javadoc:
>`@param` notifyForDescendents If `true` changes to URIs beginning with `uri`
 will also cause notifications to be sent. If `false` only changes to the exact URI
 specified by `uri` will cause notifications to be sent. If `true`, any URI values
 at or below the specified URI will also trigger a match.

JakeWharton:

PR #147, opened 2015-03-11
Add OnSubscribeQuery (a variation of OnSubscribeCursor)

wuman:
`ContentObservable.fromCursor()` is useful when we get hold of a `Cursor` instance from some asynchronous query already.  However, in some cases it might be useful to make the query asynchronously in the same observable creation function.

This pull request adds another herlper method `ContentObservable.fromQuery()` which takes as input parameters much like those of `AsyncQueryHandler.startQuery()` and returns an `Observable` emitting Cursor resulting from the query.

### DIFF
--- a/rxandroid/src/main/java/rx/android/content/ContentObservable.java
+++ b/rxandroid/src/main/java/rx/android/content/ContentObservable.java
@@ -13,11 +13,13 @@
  */
 package rx.android.content;
 
+import android.content.ContentResolver;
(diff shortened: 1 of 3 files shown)


artem-zinnatullin:
`RxAndroid` will become too large and complex if all features like this will be implemented in single library...

Take a look at [SqlBrite](https://github.com/square/sqlbrite)

JakeWharton:

PR #152, opened 2015-04-13
LifecycleObservable.bindActivityLifecycle without inheritance

nhachicha:
A companion method to `LifecycleObservable.bindActivityLifecycle` that uses activity lifecycle callbacks provided in API 14 to avoid inheritance


rxandroid/src/main/java/rx/android/lifecycle/LifecycleHelper.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * This instance lives within the scope of an {@code Activity} in order to forward Lifecycle events
+ * to {@link rx.subjects.BehaviorSubject}
+ *
+ * Thread safe, since it's <i>Thread-confined</i> to the main Thread.
+ *
+ * @author Nabil Hachicha.
+ */
+class LifecycleHelper implements Application.ActivityLifecycleCallbacks {
+    private final Activity activityToMonitor;
```
dlew: I've always thought it would be cool to have the `Subject` output all events, then use `Observable.filter()` to limit which ones you care about. Would get rid of all the `==` checks throughout the rest of this class.

rxandroid/src/main/java/rx/android/lifecycle/LifecycleHelper.java
```diff
@@ -0,0 +1,92 @@
...
+    private final BehaviorSubject<LifecycleEvent> lifecycleSubject;
+    /**
+     * Help track if {@link #onActivityCreated(Activity, Bundle)} was called
+     */
+    private boolean isCreateCalled = false;
```
dlew: It feels like there ought to be a more consistent way to track this. At the very least you could use `BehaviorSubject.hasValue()` instead of these booleans.

rxandroid/src/main/java/rx/android/lifecycle/LifecycleHelper.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Help track if {@link #onActivityStarted(Activity)} was called
+     */
+    private boolean isStarted = false;
+
+    public LifecycleHelper(Activity instance, BehaviorSubject<LifecycleEvent> lifecycle) {
```
dlew: It seems odd to me that you provide the lifecycle object when `LifecycleHelper` is the one that manages it. Why not have `LifecycleHelper` create the `BehaviorSubject`, then you can retrieve it after construction?

rxandroid/src/main/java/rx/android/lifecycle/LifecycleHelper.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * This instance lives within the scope of an {@code Activity} in order to forward Lifecycle events
+ * to {@link rx.subjects.BehaviorSubject}
+ *
+ * Thread safe, since it's <i>Thread-confined</i> to the main Thread.
+ *
+ * @author Nabil Hachicha.
+ */
+class LifecycleHelper implements Application.ActivityLifecycleCallbacks {
```
dlew: I'm not a big fan of this name; "helper" is generic, but really this is just an implementation of `ActivityLifecycleCallbacks`. Maybe something like `ActivityLifecycleCallbacksObservable`?

dlew: Actually, on that note, this could actually be an `OnSubscribe` (in the same way that a lot of the observables in RxAndroid are `OnSubscribes`). Then it would get rid of the need for an internal `Subject`.

rxandroid/src/main/java/rx/android/lifecycle/LifecycleObservable.java
```diff
@@ -72,6 +77,30 @@ public Boolean call(LifecycleEvent lifecycleEvent) {
...
+     *
+     * @param activity the activity we want to monitor lifecycle sequence for
+     * @param source   the source sequence
+     */
+
+    @TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+    public static <T> Observable<T> bindActivityLifecycle(Activity activity, Observable<T> source) {
+        // Make sure we're running on ICS or higher to use ActivityLifecycleCallbacks
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.ICE_CREAM_SANDWICH) {
+            throw new IllegalStateException ("This method is only available on API >= 14");
```
dlew: I don't believe this library supports minSdkVer < 14 anyways. Could be wrong though.

rxandroid/src/main/java/rx/android/lifecycle/LifecycleHelper.java
```diff
@@ -0,0 +1,92 @@
...
+
+import rx.subjects.BehaviorSubject;
+
+/**
+ * This instance lives within the scope of an {@code Activity} in order to forward Lifecycle events
+ * to {@link rx.subjects.BehaviorSubject}
+ *
+ * Thread safe, since it's <i>Thread-confined</i> to the main Thread.
+ *
+ * @author Nabil Hachicha.
```
JakeWharton: Remove. This library is free of individual authors and individual copyright owners.

rxandroid/src/main/java/rx/android/lifecycle/LifecycleHelper.java
JakeWharton: Yes pls. To all the things.

rxandroid/src/main/java/rx/android/lifecycle/LifecycleHelper.java
```diff
@@ -0,0 +1,92 @@
...
+        if (activity == activityToMonitor) {
+            lifecycleSubject.onNext(LifecycleEvent.CREATE);
+            isCreateCalled = true;
+        }
+    }
+
+    @Override
+    public void onActivityStarted(Activity activity) {
+        if (activity == activityToMonitor) {
+            if (!isCreateCalled) {//client bound onStart, sending missed event
```
JakeWharton: I question this behavior. If I bind in onStart I would expect to have missed onCreate. This is an application of lifecycle events which would be a hot observable. If you aren't around to observe the events, you don't get them.

rxandroid/src/main/java/rx/android/lifecycle/LifecycleObservable.java
```diff
@@ -72,6 +77,30 @@ public Boolean call(LifecycleEvent lifecycleEvent) {
...
+
+        } else {
```
JakeWharton: You don't need `else` after a `throw`.

rxandroid/src/main/java/rx/android/lifecycle/LifecycleHelper.java
kungfurorykelly: I second this.

rxandroid/src/main/java/rx/android/lifecycle/LifecycleHelper.java
```diff
@@ -0,0 +1,92 @@
...
+ * This instance lives within the scope of an {@code Activity} in order to forward Lifecycle events
+ * to {@link rx.subjects.BehaviorSubject}
+ *
+ * Thread safe, since it's <i>Thread-confined</i> to the main Thread.
+ *
+ * @author Nabil Hachicha.
+ */
+class LifecycleHelper implements Application.ActivityLifecycleCallbacks {
+    private final Activity activityToMonitor;
+    private final BehaviorSubject<LifecycleEvent> lifecycleSubject;
```
kungfurorykelly: Why is this a BehaviorSubject. I'm relatively new to RX but won't this begin by emitting the last lifecycle event received and then any other lifecycle events it receives. Is that the behaviour you want? Is this to catch the case when someone subscribes after onDestroy?

sample-app/src/main/java/rx/android/samples/LifecycleObservableActivityComposition.java
```diff
@@ -0,0 +1,83 @@
...
+        button.setText("Click Me!");
+        setContentView(button);
+    }
+
+    @Override
+    protected void onStart() {
+        super.onStart();
+
+        subscription =
+                LifecycleObservable.bindActivityLifecycle(this, ViewObservable.clicks(button))
```
kungfurorykelly: After reading http://blog.danlew.net/2015/03/02/dont-break-the-chain/ would it be Nicer to do something like?
```
subscription = ViewObservable.clicks(button) 
    .compose(bindActivityLifecycle(this)) 
```

iNoles: Transformer<T, R> will return R. With Java's 7 Type Inference, R will be Object. You have to do 

```java LifecycleObservable.<Specific Object Class>bindActivityLifecycle(this)```

http://docs.oracle.com/javase/tutorial/java/generics/genTypeInference.html

Comment (kungfurorykelly) [inline sample-app/src/main/java/rx/android/samples/LifecycleObservableActivityComposition.java]: Ok thanks a lot! Sorry I've been using [retrolamda](https://github.com/orfjackal/retrolambda) and java 8 with Rx.

**Comment by nhachicha** [comment]
Hi @dlew @JakeWharton 
I updated the  PR
-  Using  a `filter`  for https://github.com/ReactiveX/RxAndroid/pull/152#discussion-diff-28239853
- Removing boolean https://github.com/ReactiveX/RxAndroid/pull/152#discussion_r28240019
- Using `OnSubscribes ` https://github.com/ReactiveX/RxAndroid/pull/152#discussion_r28240354
- Removed author https://github.com/ReactiveX/RxAndroid/pull/152#discussion_r28244467 & useless `else` https://github.com/ReactiveX/RxAndroid/pull/152#discussion_r28245293
- Fixed behaviour in https://github.com/ReactiveX/RxAndroid/pull/152#discussion_r28244953
We can't rely on `Application.ActivityLifecycleCallbacks` to detect precisely the _first event_ ex:
```java
    @Override
    protected void onResume() {
        subscription =
                LifecycleObservable.bindActivityLifecycle(this,
                        ViewObservable.clicks(button),
                        LifecycleEvent.START)
                        .subscribe(new Action1<OnClickEvent>() {
                            @Override
                            public void call(OnClickEvent onClickEvent) {
                                Toast.makeText(LifecycleObservableActivityComposition.this,
                                        "Clicked button!",
                                        Toast.LENGTH_SHORT)
                                        .show();
                            }
                        });
         // LifecycleCallbacks will catch this event
        super.onResume();
    }

```
In this case the first event is `onResume` which is correct (no need to use booleans or `BehaviorSubject.hasValue()` to adjust the first event). Now imagine this use case

```java
    @Override
    protected void onResume() {
         // LifecycleCallbacks will not catch this event
        super.onResume();
        subscription =
                LifecycleObservable.bindActivityLifecycle(this,
                        ViewObservable.clicks(button),
                        LifecycleEvent.START)
                        .subscribe(new Action1<OnClickEvent>() {
                            @Override
                            public void call(OnClickEvent onClickEvent) {
                                Toast.makeText(LifecycleObservableActivityComposition.this,
                                        "Clicked button!",
                                        Toast.LENGTH_SHORT)
                                        .show();
                            }
                        });
    }
```
From a `LifecycleCallbacks` we don't know what was the first event the user used to bind (could be `onCreate` could be `onStart`) in order to apply a _correction_ (i.e emitting previous events).

This is why the method accept this now __as a parameter__ from the user (like `LifecycleObservable#bindUntilLifecycleEvent()`).

- Add unit tests

rxandroid/src/main/java/rx/android/lifecycle/OnSubscribeActivityLifecycleCallbacks.java
```diff
@@ -0,0 +1,156 @@
...
+    public void call(Subscriber<? super LifecycleEvent> subscriber) {
+        this.subscriber = subscriber;
+    }
+
+    @Override
+    public void onActivityCreated(final Activity activity, Bundle bundle) {
+        Observable
+                .just(activity)
+                .filter(sameInstance)
+                .subscribe(sendCreateEvent);
```
JakeWharton: Using an `Observable` in all these callbacks is wasteful in both allocation and performance. I don't see any advantage over
```java
if (activity == activityToMonitor) {
  subscriber.onNext(LifecycleEvent.CREATE);
}
```
which is allocation-free.

rxandroid/src/main/java/rx/android/lifecycle/LifecycleObservable.java
```diff
@@ -72,6 +76,50 @@ public Boolean call(LifecycleEvent lifecycleEvent) {
...
+     *          LifecycleEvent.START)
+     *         .subscribe(...);
+     *   }
+     *  }
+     * }
+     * </pre>
+     *
+     * @param activity the activity we want to monitor lifecycle sequence for
+     * @param source   the source sequence
+     * @param bindEvent the binding {@link LifecycleEvent} associated with the <code>source</code> Observable.
```
JakeWharton: Why do I pass in the lifecycle event in which I'm doing the binding instead of passing in the lifecycle event which should terminate the subscription? The latter seems much more intuitive and declarative of intent.

rxandroid/src/main/java/rx/android/lifecycle/OnSubscribeActivityLifecycleCallbacks.java
nhachicha: Spot on! changed to `bindActivityUntilLifecycle`

rxandroid/src/main/java/rx/android/lifecycle/LifecycleObservable.java
nhachicha: make sense, refactored.

nhachicha:
@JakeWharton updated PR
Cheers,

nhachicha:
Ready/Waiting for review @JakeWharton @dlew  :)

JakeWharton:
As part of #172 LifecycleObservable has been removed from future releases. While something like it may come back (and not favoring inheritance) we are focused on making a 1.0 of a core that's fundamental to all users of RxJava on Android.

PR #156, opened 2015-04-20
Update gradle versions.

IgorGanapolsky:


JakeWharton:
This was done in a separate PR. But thanks!

PR #159, opened 2015-04-22
Added RxService & RxIntentService

SalomonBrys:
This is creates `LifecycleObservable.bindServiceLifecycle` that works the exact same way as `LifecycleObservable.bindActivityLifecycle`.

This allows subscriber or Observables that are bound to the service to be unsubscribed silently when the service is destroyed.


JakeWharton:
As part of #172 we are only including things in the core that are applicable to almost every user. Additionally, LifecycleObservable has been removed for a future release in favor of explicit unsubscription or implementing lifecycle notification yourself.

Issue #158, opened 2015-04-22
ConcurrentModificationException when ViewObservable.clicks().take(1)


shinyaohira:
I would like to do something just one time when users tap a button. Here is my code:

```java
ViewObservable.clicks(button)
                .take(1)
                .subscribe(/* do something */);
```

but I got ConcurrentModificationException.

```
java.util.ConcurrentModificationException
            at java.util.ArrayList$ArrayListIterator.next(ArrayList.java:573)
            at rx.android.view.OnSubscribeViewClick$CompositeOnClickListener.onClick(OnSubscribeViewClick.java:78)
            at android.view.View.performClick(View.java:4780)
            at android.view.View$PerformClick.run(View.java:19866)
```

iyotetsuya:
try this, it works for me.
```java 
ViewObservable.clicks(button)
    .observeOn(AndroidSchedulers.mainThread())
    .take(1)
    .subscribe(/* do something */);
```

JakeWharton:
This is fixed in https://github.com/JakeWharton/RxBinding but per #172 this won't be going into RxAndroid core. We have removed the view binding observables for the next release.

PR #162, opened 2015-04-26
Fixed concurrent modifications in UI components when subscribing taken events

jmatsu:
This fixes ConcurrentModificationException which is thrown by Observables that invoked Observable#take(int).

Relevant issue is #158 .

### DIFF
--- a/rxandroid/src/main/java/rx/android/view/OnSubscribeViewClick.java
+++ b/rxandroid/src/main/java/rx/android/view/OnSubscribeViewClick.java
@@ -25,6 +25,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.WeakHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 final class OnSubscribeViewClick implements Observable.OnSubscribe<OnClickEvent> {
     private final boolean emitInitialValue;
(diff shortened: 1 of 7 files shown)


rxandroid/src/main/java/rx/android/view/OnSubscribeViewClick.java
```diff
@@ -75,7 +75,7 @@ public boolean removeOnClickListener(final View.OnClickListener listener) {
 
         @Override
         public void onClick(final View view) {
-            for (final View.OnClickListener listener : listeners) {
+            for (final View.OnClickListener listener: new ArrayList<View.OnClickListener>(listeners)) {
```
JakeWharton: This is extremely wasteful. Try a `CopyOnWriteArrayList` for `listeners` instead?

jmatsu: Thanks for the reasonable comment. 
I modified the codes to match it, and pushed them.

ronshapiro:
Did you find out when/how the list of listeners is being modified from a background thread?

jmatsu:
I did.
This is not caused by modifications from two or more threads.

In this case, a taken (and completed) observer will try to unsubscribe events on UiThread after invoking onNext(View).
But List#remove via unsubscribing is sequentially executed because onNext(View) is called from UiThread.

So the solution is using CopyOnWriteArrayList or to create new list, I think.

rxandroid/src/main/java/rx/android/view/OnSubscribeViewClick.java
```diff
@@ -75,7 +76,7 @@ public boolean removeOnClickListener(final View.OnClickListener listener) {
 
         @Override
         public void onClick(final View view) {
-            for (final View.OnClickListener listener : listeners) {
+            for (final View.OnClickListener listener: listeners) {
```
ronshapiro: Can you revert all of these small whitespace changes, unnecessary usages of `final`?

jmatsu: @ronshapiro I see. Minimized diffs.

passsy:
+1

JakeWharton:
This is fixed in https://github.com/JakeWharton/RxBinding but per #172 this won't be going into RxAndroid core. We have removed the view binding observables for the next release.

PR #166, opened 2015-05-11
Adding OnFocusChange operator

alcarvalho:
I have implemented a couple of operators for a project and was wondering if you're interested on PRs for these. This is the first one. Also to check if this is the way to go.

With them, I also think it would be a good idea to extract generic CompositeListener & CachedListerener for all the cases that make sense. With this new addition we have 3 versions of them on inner classes, already.


JakeWharton:
This is already in https://github.com/JakeWharton/RxBinding but per #172 this won't be going into RxAndroid core. We have removed the view binding observables for the next release.

Issue #172, opened 2015-06-09
Library Modularization via Specialization


JakeWharton:
In an effort to stave off becoming either obsolete or a dumping ground (or both), a few of us have been talking through other channels about how best to approach the future of this library.

Everyone agrees that the current project is a dumping ground of utilities, operators, and helpers that were once useful to one or more people when working with RxJava on Android. These were contributed by a wide variety of people with little thought to what this library's identity would be—they just wanted to share useful code.

Android is an entire platform with vast APIs. Adapting or shimming all of its APIs with RxJava will result in a large, slow moving monolithic library. It would be very hard to dictate policy on naming, reference holding, and behavior across such a large adaptation. And finally, the ways with which developers interact with this large API surface area differs based on use-case, history, and opinion.

It is infeasible to build a single RxAndroid library.

This should come as no surprise as the repo currently doesn't even house a single RxAndroid library. The ill-defined 'framework' module splits functionality in an pre-emptive realization of all the points from above.

**I am proposing a from-scratch, modularization of the library into re-usable and composeable modules.** A few other notable contributors are in support of this and I'll let them voice their support, suggestions, and/or issues individually.

The APIs in RxAndroid which are fundamental to all RxJava use on the platform deals with threading. Android's hallowed main thread and the related handler-scheduler adapter are fundamental to all RxJava use in a non-trivial capacity. These will be part of the 'core' module.

The newly added plugin hooks are also something foundational with which other modules would leverage functionality. They will also be present in the 'core' module.

Finally, internal static helpers which will be used in all modules (such as the current preconditions/assertions classes) will be in the 'core' to reduce needless duplication.

The second subject on which I want to dedicate a module is binding to and from the UI controls which make up the `android.view.*` and `android.widget.*` packages. This 'ui' module will feature observable factories for listening to UI changes and action factories for subscribing UI properties.

The scope of this second module is too large for this issue as there are a lot of considerations to be made around naming, threading, reference keeping, etc. so I will create another issue for that once consensus is reached.

I'll try to preempt answers to some questions below:

**This seems drastic. Is this really needed?**

It has become very clear to the few of us that have been having conversations that this project is dying. If we do not undertake an effort to make drastic changes such that applications and libraries can confidently depend on it in a way that's useful to everyone then it will simply stagnate into uselessness. 

The current offering covers Android APIs randomly, has varying implementations of correctness and efficient, and without a strong idea of an identity no one know what should go in.

**Will this be done in an API-compatible way?**

[...]

PR #180, opened 2015-06-20
Support query cursor on subscriber's thread

vladlichonos:
Extended ContentObservable to be able to postpone query of cursor and get it asynchronously


JakeWharton:

vladlichonos:
Got it, agree!

PR #188, opened 2015-06-30
Expose scheduler factory for Handler on HandlerScheduler.

JakeWharton:
This also renames HandlerThreadScheduler to HandlerSchduler as a HandlerThread is another concept.


rxandroid/src/main/java/rx/android/schedulers/AndroidSchedulers.java
```diff
@@ -18,27 +18,16 @@
...
-    }
+            new HandlerScheduler(new Handler(Looper.getMainLooper()));
 
-    /**
-     * {@link Scheduler} which will execute actions on the Android UI thread.
-     */
+    /** A {@link Scheduler} which executes actions on the Android UI thread. */
     public static Scheduler mainThread() {
         Scheduler scheduler =
                 RxAndroidPlugins.getInstance().getSchedulersHook().getMainThreadScheduler();
```
artem-zinnatullin: Makes me sad :( 

For just resolving `mainThread` scheduler we have to do calls to 3 virtual methods and then execute conditional check. This hook probably will be used by 1% of library users and it will give performance reduction (small, but, we are on mobile devices) for 99% other users. 

May be we can offer something like "RxAndroidPlugins.enableHooks()" which will change some boolean variable and we will check it (via static getter or whatever) before trying to resolve things via hooks? 

I understand that this is not clean solution, tho it's more likely "just saying", but I just wanted to say that RxJava already adds some overhead in compare to regular async code, so it would be nice to minimize such overhead in RxAndroid.

rxandroid/src/test/java/rx/android/schedulers/HandlerSchedulerTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Test
+    public void shouldScheduleDelayedActionOnHandlerThread() {
+        Handler handler = mock(Handler.class);
+        @SuppressWarnings("unchecked")
+        Action0 action = mock(Action0.class);
+
+        Scheduler scheduler = HandlerScheduler.create(handler);
+        Worker inner = scheduler.createWorker();
+        inner.schedule(action, 1L, TimeUnit.SECONDS);
```
artem-zinnatullin: May be static imports for `TimeUnit` values here and below?

rxandroid/src/main/java/rx/android/schedulers/AndroidSchedulers.java
dlew: I disagree that only 1% of library users will use the hook. Anyone who wants to do testing with RxAndroid will want this hook.

rxandroid/src/main/java/rx/android/schedulers/HandlerScheduler.java
```diff
@@ -24,30 +24,32 @@
 import rx.subscriptions.Subscriptions;
 import android.os.Handler;
 
-class HandlerThreadScheduler extends Scheduler {
+/** A {@link Scheduler} backed by a {@link Handler}. */
+public final class HandlerScheduler extends Scheduler {
+    /** Creates a {@link Scheduler} which uses {@code handler} to execute actions. */
+    public static HandlerScheduler create(Handler handler) {
```
artem-zinnatullin: What about `from(handler)` instead of `create(handler)` for consistency with `Schedulers.from(executor)`?

Also, users may expect that `AndroidSchedulers` will provide method `create(handler)`, because `Handler` is Android specific thing and `mainThread()` (which is based on `Handler`) lives in `AndroidSchedulers`, as far as I see, all RxJava's schedulers provided via single class — `Schedulers` so for consistency we can provide all RxAndroid's schedulers via `AndroidSchedulers`. 

Maybe we can move `create(handler)` to `AndroidSchedulers` or move `mainThread()` to `HandlerScheduler` for more obvious code structure?

JakeWharton: I'll rename to `from`. Disagree with everything else.

rxandroid/src/main/java/rx/android/schedulers/AndroidSchedulers.java
artem-zinnatullin: We would live in way more better world if at least 1% of Android apps would
have tests :)

For example, we do test our code (Unit and Integration) and we don't need
these schedulers hooks.. I'm not suggesting to remove them, just noticed
that most popular thing in RxAndroid requires several methods calls.
30 июня 2015 г. 9:15 PM пользователь "Daniel Lew" <notifications@github.com>
написал:

artem-zinnatullin:
:+1:

dlew:
LGTM

PR #189, opened 2015-06-30
Use Android build system, generate AARs

dlew:
Also updated to latest & greatest dependencies while I was at it.

This PR gets rid of [gradle-rxjava-project-plugin](https://github.com/nebula-plugins/gradle-rxjava-project-plugin) because it's incompatible with the Android build plugins. It's actually a smorgasbord of plugins, some of which seem to be geared only towards the Java plugin.

I think we'll end up having to import the individual plugins as necessary for release. I'm thinking we can handle that in parallel with other development; having this library work easily in Android Studio should really help out development.

### DIFF
--- a/sample-app/build.gradle
+++ b/sample-app/build.gradle
@@ -1,28 +1,21 @@
-buildscript {
-    repositories { jcenter() }
-    dependencies {
-        classpath 'com.android.tools.build:gradle:1.2.3'
-    }
-}
-
-repositories {
-    jcenter()
-}
-
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 22
-    buildToolsVersion "22.0.1"
+  compileSdkVersion rootProject.ext.compileSdkVersion
+  buildToolsVersion rootProject.ext.buildToolsVersion
 
-    defaultConfig {
-        minSdkVersion 14
-        targetSdkVersion 22
-        versionCode 1
-        versionName "1.0"
-    }
+  defaultConfig {
+    minSdkVersion rootProject.ext.minSdkVersion
+    targetSdkVersion rootProject.ext.targetSdkVersion
+    versionCode 1
+    versionName '1.0'
+  }
+}
+
+repositories {
+  mavenCentral()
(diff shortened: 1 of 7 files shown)


sample-app/build.gradle
```diff
@@ -1,28 +1,21 @@
...
-    compileSdkVersion 22
-    buildToolsVersion "22.0.1"
+  compileSdkVersion rootProject.ext.compileSdkVersion
+  buildToolsVersion rootProject.ext.buildToolsVersion
+
+  defaultConfig {
+    minSdkVersion rootProject.ext.minSdkVersion
+    targetSdkVersion rootProject.ext.targetSdkVersion
+    versionCode 1
+    versionName "1.0"
```
artem-zinnatullin: `versionName` value can be set with single quotes `'1.0'` (saves some nanoseconds on string parsing -> faster build :smile:)

settings.gradle
```diff
@@ -1,4 +1,2 @@
-rootProject.name='rxandroid-root'
```
JakeWharton: Keep this. Otherwise you'll get `rxandroid-rxandroid` as a module name for the nested project.

artem-zinnatullin:
:+1: for Android Studio support, probably you also need to add androidJavadoc and androidSources tasks here or in separate PR, because now they are not generated, as far as I see.

austynmahoney:
I opened this in Android Studio 1.3 Beta 1. I had the SDK tools fully updated and the m2 repos set up as I would normally. Sample runs, and project builds.

It would be nice if a named aar was created instead of the default `library.aar`. I can add this in another PR once this is merged if you like.

:+1:

JakeWharton:
The files get named appropriately by default:
```
$ ls rxandroid/build/outputs/aar/
total 32
-rw-r--r--  1 jw  jw   7.4K Jun 30 16:23 rxandroid-debug.aar
-rw-r--r--  1 jw  jw   7.3K Jun 30 16:23 rxandroid-release.aar
```

PR #190, opened 2015-06-30
minSdkVersion=9

### LINKED ISSUE #68: Minimum SDK Version
I was curious if RxAndroid has an official stance on the minimum supported SDK

I noticed recently that `View.addOnAttachStateChangeListener()` was used, which was added in 12. If we're allowing that, I propose that we might as well go to 15 (since no one is on honeycomb anyways).

JakeWharton:
Closes #68.

### DIFF
--- a/rxandroid/src/main/AndroidManifest.xml
+++ b/rxandroid/src/main/AndroidManifest.xml
@@ -12,4 +12,9 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
-<manifest package="rx.android" />
+<manifest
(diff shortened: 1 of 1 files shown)


dlew:
We should just modify `rootProject.ext.minSdkVersion` for this.

roman-mazur:
It will be overridden by gradle
https://github.com/ReactiveX/RxAndroid/blob/0.x/rxandroid/build.gradle#L8

JakeWharton:
@dlew nice. didn't see you set those. reverted.

PR #193, opened 2015-07-01
Remove targetSdkVersion from library.

JakeWharton:
This value participates in manifest merging downstream and while I'm a strong advocate for always targetting the latest, we gain nothing by specifying it and will inevitably cause someone a merging error.

### DIFF
--- a/rxandroid/build.gradle
+++ b/rxandroid/build.gradle
@@ -6,7 +6,6 @@ android {
 
     defaultConfig {
         minSdkVersion rootProject.ext.minSdkVersion
-        targetSdkVersion rootProject.ext.targetSdkVersion
     }
 
     compileOptions {


roman-mazur:
But isn't it possible to resolve the error with `tools:` directives in the app manifest?
Omitting target sdk version might potentially have implications on how lint works for this library.

JakeWharton:
I'm not going to force people to do that as that is exceedingly annoying of a library to do. Unless you have a concrete example of something that will change in lint that is essential to the library, the hypothetical prospect of something changing is not enough to stop me wanting this removed.

roman-mazur:
No real examples for now.

austynmahoney:
Another hypothetical unfortunately, but just want to bring it up for discussion. This could have implications when a developer changes their target SDK. If the SDK changes enough, but RxAndroid has not been updated, it may break. Setting a specific SDK level for the library should guarantee it will work. Android has in the past left backwards compatibility in depending on the `targetSdkVersion` set, and removed it when targeting latest.

dlew:
:+1: to removing from me.

PR #196, opened 2015-07-02
Publishing releases draft

dlew:
An attempt at getting us back to a releasable state.

The bintray/maven configuration values are derived from the pre-existing releases + what defaults can be gleaned from the previously used [gradle-rxjava-project-plugin](https://github.com/nebula-plugins/gradle-rxjava-project-plugin).

This is a draft for a few reasons:

1. I'm not sure that the values are correct.
1. I don't have API access so I can't directly test whether it actually works.
1. `bintray` has `dryRun = true` at the moment, so I won't accidentally break anything.

Suggestions on how to improve/fix are more than welcome!

### DIFF
--- a/gradle/artifacts.gradle
+++ b/gradle/artifacts.gradle
@@ -1,0 +1,64 @@
+if (project.plugins.hasPlugin('com.android.library')) {
+    android.libraryVariants.all { variant ->
+        Task javadocTask = task("generate${variant.name.capitalize()}Javadoc", type: Javadoc) {
+            group = 'artifact'
+            description "Generates Javadoc for $variant.name"
+
+            // Source files from the variant
+            source = variant.javaCompiler.source
+
+            // Classpath from the variant + android.jar
+            String androidJar = "${android.sdkDirectory}/platforms/${android.compileSdkVersion}/android.jar"
+            classpath = variant.javaCompiler.classpath + files(androidJar)
+
+            // The Android online reference doesn't include package-list, so we have to use the local one
+            String packageListRef = "${android.sdkDirectory}/docs/reference/"
+            options.linksOffline 'http://d.android.com/reference/', packageListRef
+
+            // Additional links for any RxJava references
+            options.links 'http://reactivex.io/RxJava/javadoc/'
+
+            // Exclude generated files
+            exclude '**/BuildConfig.java'
+            exclude '**/R.java'
+
+            // Output to a unique javadoc folder per variant
+            destinationDir = new File(project.docsDir, "javadoc-$variant.name")
+        }
+
+        // For official releasese, don't prefix the name so the artifact is published correctly
+        // (Can't seem to modify it for publishing, for whatever reason...)
+        String classifierPrefix = (variant.name == 'release') ? '' : "$variant.name-"
+
+        Task javadocJarTask = task("generate${variant.name.capitalize()}JavadocJar", type: Jar, dependsOn: javadocTask) {
+            group = 'artifact'
+            description = "Generates Javadoc jar for $variant.name"
+
+            classifier = "${classifierPrefix}javadoc"
+            from javadocTask.destinationDir
+        }
+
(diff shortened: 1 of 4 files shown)


gradle/artifacts.gradle
```diff
@@ -0,0 +1,64 @@
...
+            classifier = "${classifierPrefix}javadoc"
+            from javadocTask.destinationDir
+        }
+
+        Task sourcesJarTask = task("generate${variant.name.capitalize()}SourcesJar", type: Jar) {
+            group = 'artifact'
+            description = "Generates sources jar for $variant.name"
+
+            classifier = "${classifierPrefix}sources"
+            from variant.javaCompiler.source
```
roman-mazur: Should not this be `javaCompile.source`?

build.gradle
```diff
@@ -13,6 +13,21 @@ ext {
...
+    POM_DESCRIPTION= "RxAndroid"
+
+    POM_URL="https://github.com/ReactiveX/RxAndroid"
+    POM_SCM_URL="https://github.com/ReactiveX/RxAndroid"
+    POM_SCM_CONNECTION="scm:git:https://github.com/ReactiveX/RxAndroid.git"
+    POM_SCM_DEV_CONNECTION="scm:git:git@github.com:ReactiveX/RxAndroid.git"
+
+    POM_LICENCE_NAME="The Apache Software License, Version 2.0"
+    POM_LICENCE_URL="http://www.apache.org/licenses/LICENSE-2.0.txt"
+    POM_LICENCE_DIST="repo"
```
mttkay: could we pull these into a `.properties` file?

gradle/artifacts.gradle
dlew: No. `variant.java.compile` is deprecated. From [BaseVariant.java](https://android.googlesource.com/platform/tools/base/+/master/build-system/gradle-core/src/main/groovy/com/android/build/gradle/api/BaseVariant.java):

```java
/**
 * Returns the Java Compilation task if javac was configured to compile the source files.
 * @deprecated prefer {@link #getJavaCompiler} which always return the java compiler task
 * irrespective of which tool chain (javac or jack) used.
 */
@Nullable
@Deprecated
JavaCompile getJavaCompile() throws IllegalStateException;

/**
 * Returns the Java Compiler task which can be either javac or jack depending on the project
 * configuration.
 */
@NonNull
AbstractCompile getJavaCompiler();
```

build.gradle
dlew: Could call - stuffed them into `gradle.properties`.

gradle/artifacts.gradle
roman-mazur: Ah, indeed. Jack is coming :)

rxandroid/build.gradle
```diff
@@ -26,3 +41,76 @@ dependencies {
...
+
+// Configure gradle-bintray-plugin
+bintray {
+    configurations = ['archives']
+
+    // TODO: DELETE ME WHEN READY!
+    dryRun = true
+
+    pkg {
+        repo = 'RxJava'
```
JakeWharton: Why is this RxJava and not RxAndroid?

dlew: As far as I could tell, this was the data sent by the rxjava nebula plugin: https://github.com/nebula-plugins/gradle-rxjava-project-plugin/blob/gradle-2.2/src/main/groovy/nebula/plugin/rxjavaproject/RxJavaPublishingPlugin.groovy

JakeWharton:
Any reason not to merge?

dlew:
I'm on vacation and can't fix anything until Sunday, if that's a reason.

dlew:
I'm back now.

I'm still not confident the code is correct (since I could never actually test it out); only someone with the keys to the kingdom could do that I think. But I'm fairly sure this is a step in the right direction.

dlew:
I've added artifactory support for snapshots.

Again, without credentials I can't test the code directly - but theoretically if this were merged then travis would upload a snapshot.

PR #199, opened 2015-07-17
Lifecycle Subscriptions

laynepenney:
This merge request adds some support for automatic subscription management for UI (and potentially other) subscribers. I need to write some tests, but wanted to see if it is even helpful to anyone first. I created a separate module, but I can make any adjustments necessary.

In the primary sample:
```
subscriptions().with(neverEnding())
                .subscribeOn(HandlerScheduler.from(backgroundHandler))
                        // Observe until Stop
                .observeUntil(LifecycleEvent.STOP)
                        // Be notified on the main thread
                .observeOnMainThread()
                .subscribe(new Subscriber<String>() {
                    @Override public void onCompleted() {
                        Log.d(TAG, "onCompleted()");
                    }

                    @Override public void onError(Throwable e) {
                        Log.e(TAG, "onError()", e);
                    }

                    @Override public void onNext(String string) {
                        textView.setText(string);
                    }
                });
```

this code will automatically subscribe in in onStop(). neverEnding() is an observable that will never call oncomplete.

Thanks,
Layne


fengdai:
:+1:

JakeWharton:
Hey @laynepenney,

This looks quite comprehensive and opinionated! At this time, and as you may have seen, we just finished tearing out an Android lifecycle tie-in from this library for the forthcoming 1.0 release. There was simply too much code in the project which was only applicable to a small amount of users/user-cases and/or had opinions (correct or not) which belong in a standalone library rather than something that's trying to be the canonical base library for adapting RxJava to Android.

This is something that you might want to pursue as a standalone library to design, evolve, and test the merits of separately. I'll also note that contributor @dlew is doing the same with [RxLifecycle][1] which you might want to look to for inspiration or ideas.

We're focused on shipping the bare minimum as a 1.0 so that libraries like RxLifecycle and others can start depending on a stable RxAndroid (albeit a teensy tiny one) before we would consider bringing in more heavyweight utilities like this.

 [1]: https://github.com/trello/RxLifecycle

laynepenney:
Thanks @JakeWharton. I appreciate the feedback. I'll checkout RxLifecycle and consider something standalone as you suggested. I've been working on a network/caching library built on Rx and this was part of what I came up with to handle subscriptions attached to the Android lifecycle.

PR #207, opened 2015-08-05
Remove property-based hook configuration option.

JakeWharton:
Android restricts properties to 32 characters which means these never could be set. Additionally, using properties to control behvior on Android has little precedent so we favor explicit APIs instead. If this every changes, the behavior can be changed in a backwards compatible way whereas including them as-is would prevent future removal.

Additionally, we expose the  method as a beta API. The discussion around exposing a method like this for RxJava is ongoing.


dlew:
:+1:

PR #218, opened 2015-08-13
Upgrade RxJava from 1.0.13 to 1.0.14

vanniktech:


JakeWharton:
Just a reminder: RxAndroid will _never_ be released to track RxJava versions so if you want 1.0.14 you will have to specify its dependency explicitly in your app.

PR #220, opened 2015-08-25
Add consumer proguard rules for RxJava

### LINKED ISSUE #219: Add a proguard file for sample app.
RxJava includes some aspects that can cause problems w/ proguard: https://github.com/ReactiveX/RxJava/issues/3097

However, since it's just a JAR, it cannot itself include proguard protections. I think we should add some reasonable default protections to RxAndroid instead (even though we're not directly the source of it).

felipecsl:
Fixes #219


JakeWharton:
Can you change the commit message to something descriptive of the actual change? Like "Add consumer proguard rules for RxJava".

JakeWharton:
We also should enable proguard for the sample to ensure the rules actually work.

felipecsl:
Done. enabled on the sample as well

hzsweers:
> -keep class rx.internal.util.unsafe.** { *; }

Does this have to be this broad? This adds about 7kB vs if the whole thing were proguarded. Not too bad, but could the following cover our needs?

```
keepclassmembers class rx.internal.util.unsafe.*ArrayQueue*Field* {
   long producerIndex;
   long consumerIndex;
}
keepclassmembers class rx.internal.util.unsafe.BaseLinkedQueueProducerNodeRef {
   long producerNode;
   long consumerNode;
}
```

felipecsl:
Sample updated. @hzsweers the narrower rules worked fine. Without it, I got `NoSuchFieldException: producerIndex`. With the rules, the app works correctly on release.

Also had to add a signing config to the sample app in order to test, not sure if that's ok. I can revert that otherwise

roman-mazur:
RxJava, to my understanding, uses reflection in many places. Including non-internal packages.
Like [OperatorConcat](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OperatorConcat.java#L89).
Hence, my understanding is that we need to preserve members for all classes in `rx.**` packages.

dlew:
I think we should keep the *essentials* and that the rest should be left up to the individual consumers of Rx.

dimsuz:
@dlew in this case (keeping the essentials) some kind of note in the dedicated Proguard section would probably be a good thing. So that people would know what to do when something goes wrong for them.

hzsweers:
Applying a broad/sweeping preservation is heavy handed, the narrowest possible scope should be used. 

I think this project should be reactive (no pun intended) to other proguard issues on a case-by-case basis. A wiki page can be maintained and updated as things are reported, and those updates can go out when a release happens. Users can always update their local configurations if need be rather than wait for a release, so we don't need to be too defensive against it.

hzsweers:
@JakeWharton @dlew any plans for this PR?

dlew:
I'd like to see this merged.

I understand the point of [RxJavaProGuardRules](https://github.com/artem-zinnatullin/RxJavaProGuardRules) but I highly doubt that RxJava 1.0 is going to change the signatures for these classes.

Worst-case scenario, we just proguard both versions of it.

artem-zinnatullin:
@dlew yeah, would be nice if we'll be protected by the RxAndroid too. 

Though I am [planning](https://github.com/artem-zinnatullin/RxJavaProGuardRules/issues/4) to run RxJava's unit tests on Android emulator as part of RxJavaProGuardRules CI to check that rules cover as many possible cases as possible. 

Not sure that RxAndroid will go that way (it's not very trivial because not all RxJava tests compiles with Android SDK…).

Anyway, let's finally merge this! A lot of people affected by this and most of the even don't know about this issue until they build a release apk with enabled ProGuard.

vanniktech:
Any plans on merging this one?

sample-app/build.gradle
```diff
@@ -19,6 +19,26 @@ android {
...
+
+    signingConfigs {
+        release {
+            storeFile file("android.jks")
+            storePassword "android"
+            keyAlias "RxAndroidSample"
+            keyPassword "android"
+        }
+    }
+    buildTypes {
```
JakeWharton: Two blocks for this? How about moving `signingConfigs` up and combining these two `buildTypes` blocks.

JakeWharton:
I'm not a ProGuard user, so someone else will have to sign off on the rules. Everything else looks good.

artem-zinnatullin:
LGTM (we use this rules in production)

dlew:
Most of the rules still make sense; those `producerIndex`, `consumerIndex`, `producerNode` and `consumerNode` fields are being used via `UnsafeAccess`.

The only thing I'd do is change `-dontwarn sun.misc.**` to `-dontwarn sun.misc.Unsafe`, since there's no need to be so general.

artem-zinnatullin:
The only thing I'm afraid of is that as soon as we ship these rules as part
of RxAndroid we'll have to release new versions each time RxJava will be
released with another reflection lookups, that'll make things more
complicated.

Maybe we can discuss it with RxJava maintainers and publish these rules as
a separate artifact of the RxJava project? If I'll finally finish "Run
tests on Android during the CI", we'll be able to run tests on proguarded
code and catch most of the problems during regular RxJava development
cycle.

On Mon, Nov 16, 2015, 19:51 Daniel Lew <notifications@github.com> wrote:

> Most of the rules still make sense; those producerIndex, consumerIndex,
> producerNode and consumerNode fields are being used via UnsafeAccess.
>
> The only thing I'd do is change -dontwarn sun.misc.** to -dontwarn
> sun.misc.Unsafe, since there's no need to be so general.
>
> —
> <https://github.com/ReactiveX/RxAndroid/pull/220#issuecomment-157094853>.
>
-- 

@artem_zin

vanniktech:
LGTM we're also using the same rules here for Rx

felipecsl:
PR updated, commits squashed :shipit:

hzsweers:
@artem-zinnatullin That point has been addressed with you elsewhere. To not ship for fear of changes would mean we never ship. Something here is better than nothing, and seeing as this is the first time this has come up in the ~year or so that this has been active, I think we can err on the side of optimism that this won't be an oft-recurring issue. 

The good thing about proguard rules is that if they do change, they can be updated/maintained in a wiki or on the repo itself and people can manually copy them into their projects until the next release. It is not an actual blocker for using the library.

As @felipecsl said, :shipit:

artem-zinnatullin:
>The good thing about proguard rules is that if they do change, they can be updated/maintained in a wiki or on the repo itself and people can manually copy them into their projects until the next release. 

That's what I call "that'll make things more complicated." :smile: (both for users and RxAndroid maintainers) 

I mean, @JakeWharton always say:

>As a reminder: we will never make releases to track RxJava versions and the dependency should always be specified explicitly in consuming projects.

And that's a good point for various reasons. But as soon as we merge proguard rules — this principle will be broken.

hzsweers:
That's a dramatic way of looking at it. It's no different than when library
maintainers had to keep their README up to date with proguard rules. Now
libraries can package them in for convenience. Worst case: people can
include the rules themselves as the community reports them and the library
accounts for them in a future version. There would literally be nothing
stopping them from using any version of the library with other versions of
RxJava because proguard rules are not that big of a deal.

desseim:
Please note: It seems the current rules are based on [a comment of mine](https://github.com/ReactiveX/RxJava/issues/3097#issuecomment-135125272) in a related issue, but the rules relating to the preservation of `producerNode` and `consumerNode` in `BaseLinkedQueue` node ref classes was wrong. I guess no one noticed as these 2 classes are stripped away altogether by Proguard in most setups.
Anyway, I guess the rules in this patch, if it's to be merged, should be updated as per [this comment](https://github.com/ReactiveX/RxJava/issues/3097#issuecomment-158485012).

felipecsl:
oh thanks @desseim, good catch. I'll update the PR

hzsweers:
They weren't stripped away for me, and break pointing showed it was digging into those classes. The exception at runtime also consistent with those fields not being found. Maybe this changed in 1.0.16?

vanniktech:
As far as I remember I also never had this issue with 1.0.13 and 1.0.14

desseim:
@hzsweers looking at the code it seems to be the same with `1.0.16`.
Does your code step into `lvProducerNode()` at runtime ?

@vanniktech the proguard configuration added in this PR addresses changes introduced in `1.0.14` IIRC. You should have had problems with `1.0.14` (possibly) but not with `1.0.13`.

vanniktech:
Do we need to change anything for supporting [RxJava 1.1.0[(https://github.com/ReactiveX/RxJava/releases/tag/1.1.0)?

hzsweers:
@desseim I don't have access to the code that I discovered that with anymore, but the fields that it couldn't find were those `producerNode` and `consumerNode` fields. Keeping them explicitly per these rules resolved the problem as well.

hzsweers:
@vanniktech I don't think so. The things that introduced the problem (namely in BackpressureDrainManager) were promoted to official parts of the API, which I think makes this PR all the more important.

JakeWharton:
Because of the unforeseeable volatility in the rules RxJava will need and RxAndroid's comfort in infrequent releases we have decided not to ship these in the library. We do, however, welcome a set of rules to be placed in the sample app for people to copy/paste. I will update #219 to track that.

lestinYin:
in my build.gradle is compile 'com.jakewharton.rxbinding:rxbinding:0.2.0
but when i turn on  minifyEnabled true,
and proguard write
-keep class rx.**{*;}
-keep class com.jakewharton.rxbinding.**{*;}
-dontwarn com.jakewharton.**
# rxjava
-keep class rx.schedulers.Schedulers {
    public static <methods>;
}
-keep class rx.schedulers.ImmediateScheduler {
    public <methods>;
}
-keep class rx.schedulers.TestScheduler {
    public <methods>;
}
-keep class rx.schedulers.Schedulers {
    public static ** test();
}
-keepclassmembers class rx.internal.util.unsafe.*ArrayQueue*Field* {
    long producerIndex;
    long consumerIndex;
}
-keepclassmembers class rx.internal.util.unsafe.BaseLinkedQueueProducerNodeRef {
    long producerNode;
    long consumerNode;
}
-dontwarn sun.misc.Unsafe

finally when i excute my release apk show error
java.lang.IllegalStateException: Exception thrown on Scheduler.Worker thread. Add `onError` handling.
                                                       at rx.internal.schedulers.ScheduledAction.run(ScheduledAction.java:60)
                                                       at android.os.Handler.handleCallback(Handler.java:739)
                                                       at android.os.Handler.dispatchMessage(Handler.java:95)
                                                       at android.os.Looper.loop(Looper.java:148)
                                                       at android.app.ActivityThread.main(ActivityThread.java:7406)
                                                       at java.lang.reflect.Method.invoke(Native Method)
                                                       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1230)
                                                       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1120)
                                                    Caused by: rx.exceptions.OnErrorNotImplementedException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
                                                       at rx.Observable$27.onError(Observable.java:7535)
                                                       at rx.observers.SafeSubscriber._onError(SafeSubscriber.java:154)
                                                       at rx.observers.SafeSubscriber.onError(SafeSubscriber.java:111)
                                                       at rx.observers.SafeSubscriber.onNext(SafeSubscriber.java:137)
                                                       at rx.internal.operators.OperatorDoOnEach$1.onNext(OperatorDoOnEach.java:84)
                                                       at rx.internal.operators.OperatorMerge$MergeSubscriber.emitScalar(OperatorMerge.java:472)
                                                       at rx.internal.operators.OperatorMerge$MergeSubscriber.tryEmit(OperatorMerge.java:434)
                                                       at rx.internal.operators.OperatorMerge$MergeSubscriber.onNext(OperatorMerge.java:227)

could every one indicate me any errors.thank you

iNoles:
Caused by: rx.exceptions.OnErrorNotImplementedException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference

I don't think this is related to proguard.

PR #235, opened 2015-10-31
Update RxJava to 1.0.15

vanniktech:


JakeWharton:
As a reminder: we will never make releases to track RxJava versions and the dependency should always be specified explicitly in consuming projects.

artem-zinnatullin:
We can actually enforce explicit specifying RxJava dependency in the consuming projects. Changing configuration from `compile` to `provided` will exclude `dependency` from the `pom.xml`.

Let's do it?

JakeWharton:
I don't think that's a solution, and it might actually make it worse.
Another project might have a dependency on an old RxJava which would just
fulfill that role. Plus is also forces every library to explicitly depend
on RxJava when most of them don't actually care what the version is.

vanniktech:
Yup I agree with @JakeWharton. Let's make the configuration as easy as possible (by default depend on a specific version). If you a different version, just specify it.

artem-zinnatullin:
Unfortunately, it's not so easy. If you specify 1.0.14 but one of the
library specifies 1.0.15 - Gradle will use 1.0.15 and you may even not know
about it.

But ok, since other libraries do so too, fixing it in the RxAndroid won't
solve the problem.

On Sun, Nov 1, 2015, 11:42 Niklas Baudy <notifications@github.com> wrote:

> Yup I agree with @JakeWharton <https://github.com/JakeWharton>. Let's
> make the configuration as easy as possible (by default depend on a specific
> version). If you a different version, just specify it.
>
> —
> <https://github.com/ReactiveX/RxAndroid/pull/235#issuecomment-152807617>.
>
-- 

@artem_zin

JakeWharton:
We don't really care about that case as it would be fortunate behavior. If
you truly wanted .14 you could use a resolution strategy to force it.

artem-zinnatullin:
Yeah, I understand, just wanted to show that dependency management is not
such easy (hell).

On Sun, Nov 1, 2015, 15:22 Jake Wharton <notifications@github.com> wrote:

> We don't really care about that case as it would be fortunate behavior. If
> you truly wanted .14 you could use a resolution strategy to force it.
>
> On Sun, Nov 1, 2015 at 12:19 PM Artem Zinnatullin <
> notifications@github.com>
> wrote:
>
> > Unfortunately, it's not so easy. If you specify 1.0.14 but one of the
> > library specifies 1.0.15 - Gradle will use 1.0.15 and you may even not
> know
> > about it.
> >
> > But ok, since other libraries do so too, fixing it in the RxAndroid won't
> > solve the problem.
> >
> > On Sun, Nov 1, 2015, 11:42 Niklas Baudy <notifications@github.com>
> wrote:
> >
> > > Yup I agree with @JakeWharton <https://github.com/JakeWharton>. Let's
> > > make the configuration as easy as possible (by default depend on a
> > specific
> > > version). If you a different version, just specify it.
> > >
> > > —
> > > Reply to this email directly or view it on GitHub
> > > <
> https://github.com/ReactiveX/RxAndroid/pull/235#issuecomment-152807617
> > >.
> > >
> > --
> >
> > @artem_zin
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/ReactiveX/RxAndroid/pull/235#issuecomment-152822365
> >.
> >
>
> —
> <https://github.com/ReactiveX/RxAndroid/pull/235#issuecomment-152822438>.
>
-- 

@artem_zin

PR #239, opened 2015-11-11
Lazily initialize MAIN_THREAD_SCHEDULER to avoid class loading problems in unit tests

### LINKED ISSUE #238: Unit testing support for AndroidSchedulers
Currently when using `mainThread()` in a usual Unit Test, (even with an installed hook scheduler) I get
```
Caused by: java.lang.RuntimeException: Method getMainLooper in android.os.Looper not mocked. See http://g.co/androidstudio/not-mocked for details.
	at android.os.Looper.getMainLooper(Looper.java)
	at rx.android.schedulers.AndroidSchedulers.<clinit>(AndroidSchedulers.java:27) `
```
error.

So, it would be great to apply the original RxJava pattern without instantiating the defauilt scheduler during static initialization.

The curret workaround is `@SuppressStaticInitializationFor("rx.android.schedulers.AndroidSchedulers")` from PowerMock.

artem-zinnatullin:
Fixes #238.

### DIFF
--- a/rxandroid/src/main/java/rx/android/schedulers/AndroidSchedulers.java
+++ b/rxandroid/src/main/java/rx/android/schedulers/AndroidSchedulers.java
@@ -13,9 +13,10 @@
  */
 package rx.android.schedulers;
 
-import rx.Scheduler;
 import android.os.Handler;
 import android.os.Looper;
+
+import rx.Scheduler;
 import rx.android.plugins.RxAndroidPlugins;
 
 /** Android-specific Schedulers. */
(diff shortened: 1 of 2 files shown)


rxandroid/src/main/java/rx/android/schedulers/AndroidSchedulers.java
```diff
@@ -24,13 +25,17 @@ private AndroidSchedulers() {
         throw new AssertionError("No instances");
     }
 
-    private static final Scheduler MAIN_THREAD_SCHEDULER =
-            new HandlerScheduler(new Handler(Looper.getMainLooper()));
+    // See https://github.com/ReactiveX/RxAndroid/issues/238
+    // https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom
+    private static class MainThreadSchedulerHolder {
+        private static final Scheduler MAIN_THREAD_SCHEDULER =
```
JakeWharton: Make this package scoped to avoid synthetic accessor method? There's already enough method calls inside of `mainThread()` below, let's not add one more.

artem-zinnatullin: Np, fixed. No more `$access000()`.

JakeWharton:
The test change is unrelated, and seems completely unnecessary. Please split PRs into logical groupings of changes in the future.

artem-zinnatullin:
>The test change is unrelated, and seems completely unnecessary. Please split PRs into logical groupings of changes in the future.

Removing Robolectric test runner was required to see that test fails without it. I agree that I shouldn't change `Schedulers.immidiate()` to `mock()` in this PR, sorry.

JakeWharton:
Ah, got it. Forgot that was the motivation of the change. Makes sense.

PR #247, opened 2015-12-04
Adds AndroidSchedulers.currentThread()

manidesto:


rxandroid/src/main/java/rx/android/schedulers/AndroidSchedulers.java
```diff
@@ -38,4 +38,10 @@ public static Scheduler mainThread() {
                 RxAndroidPlugins.getInstance().getSchedulersHook().getMainThreadScheduler();
         return scheduler != null ? scheduler : MainThreadSchedulerHolder.MAIN_THREAD_SCHEDULER;
     }
+
+    public static Scheduler currentThread() {
```
artem-zinnatullin: Needs javadoc

artem-zinnatullin: And at least two more tests: 

1. For a `Thread` with `Looper` to check that it picks its looper and runs the code on that thread.
2. For a `Thread` without `Looper` to see how it fails with `Can't create handler inside thread that has not called Looper.prepare()`.

manidesto:
@artem-zinnatullin 
Not sure how to test that. Any suggestions?

dlew:
I'd like to see where this becomes important in practice.

It sort of seems like something you could do without a custom `Scheduler`, e.g. by querying the current `Looper` before calling `subscribeOn` or `observeOn`

JakeWharton:
I ask in the issue for this feature for some use cases which was not
fulfilled.

manidesto:
To return an observable that does work in a background thread but emits items in the same thread the consumer of the api called the function. 
Using Scheduler.immediate() or Scheduler.trampoline() may break the app if the api is called from the main thread.

Wasn't able to unit test my presenter due to issue #238 . So when moving the `observeOn()` part into the API, wanted this function so that I can call my API from any thread. Will not be much of a need if the fix for #238 is released I guess.

artem-zinnatullin:
So, if it's just for testing (I don't see any significant use cases for real usage at the moment), you should be fine with fix #238.

I'd just suggest you to make schedulers configurable in your Presenters and pass `Schedulers.trampoline()` or `Schedulers.immediate()` for tests.

@manidesto let's close this then?

manidesto:
Yeah. But please ready the next release as soon as possible.

PR #246, opened 2015-12-04
Update build scripts and sample-app

friederbluemle:
Just a few minor updates to the build scripts (Gradle wrapper `2.9`, Android plugin `1.5.0`, removed redundancies) and an updated sample-app (new app icon, fixed Lint and Javadoc warnings).

PR #256, opened 2015-12-11
Fix Javadoc warnings

friederbluemle:
Building the project will output in the following warnings:

    <...>/rx/android/schedulers/AndroidSchedulers.java:36: warning: no @return
        public static Scheduler mainThread() {
                                ^
    <...>/rx/android/schedulers/HandlerScheduler.java:30: warning: no @param for handler
        public static HandlerScheduler from(Handler handler) {
                                       ^
    <...>/rx/android/schedulers/HandlerScheduler.java:30: warning: no @return
        public static HandlerScheduler from(Handler handler) {
                                       ^
    <...>/rx/android/plugins/RxAndroidPlugins.java:58: warning: no @return
        public RxAndroidSchedulersHook getSchedulersHook() {
                                       ^
    <...>/rx/android/plugins/RxAndroidPlugins.java:74: warning: no @param for impl
        public void registerSchedulersHook(RxAndroidSchedulersHook impl) {
                    ^
    <...>/rx/android/plugins/RxAndroidSchedulersHook.java:33: warning: no @return
        public Scheduler getMainThreadScheduler() {
                         ^
    6 warnings

Given that [Javadoc](http://www.oracle.com/technetwork/articles/java/index-137868.html) is a relatively well defined standard (e.g. in regards to which tags are required and which are optional), there should be no reason to have documentation that litters the build log with warnings.

In an attempt to fix these warnings, this pull request just adds the missing `@param` and `@return` tags with short descriptions taken from the main documentation. As @JakeWharton in #246 correctly criticized, this creates some redundancies. Two other possible solutions I can think of would be to reword the main documentation in order to decrease the redundancy, or to configure the javadoc tool to ignore these issues altogether and not output warnings. Open to other suggestions as well.

### DIFF
--- a/rxandroid/src/main/java/rx/android/plugins/RxAndroidPlugins.java
+++ b/rxandroid/src/main/java/rx/android/plugins/RxAndroidPlugins.java
@@ -54,6 +54,8 @@ public final class RxAndroidPlugins {
      * Override the default by calling {@link #registerSchedulersHook(RxAndroidSchedulersHook)} or by
      * setting the property {@code rxandroid.plugin.RxAndroidSchedulersHook.implementation} with the
      * full classname to load.
+     *
+     * @return The {@link RxAndroidSchedulersHook} instance
      */
     public RxAndroidSchedulersHook getSchedulersHook() {
         if (schedulersHook.get() == null) {
@@ -68,6 +70,7 @@ public final class RxAndroidPlugins {
      * Registers an {@link RxAndroidSchedulersHook} implementation as a global override of any
      * injected or default implementations.
      *
+     * @param impl The {@link RxAndroidSchedulersHook} implementation
      * @throws IllegalStateException if called more than once or after the default was initialized
      * (if usage occurs before trying to register)
      */
--- a/rxandroid/src/main/java/rx/android/plugins/RxAndroidSchedulersHook.java
+++ b/rxandroid/src/main/java/rx/android/plugins/RxAndroidSchedulersHook.java
@@ -29,6 +29,8 @@ public class RxAndroidSchedulersHook {
      * should be used.
      * <p>
      * This instance should be or behave like a stateless singleton.
+     *
+     * @return The main thread {@link Scheduler}
      */
     public Scheduler getMainThreadScheduler() {
         return null;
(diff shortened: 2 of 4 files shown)


JakeWharton:
Yeah I'm against all of these! I actively eliminate these types of documentation from projects I work on. A well written comment almost always can provide more value and encapsulate both what the return value and expected arguments will be.

```java
/** Returns {@code base} raised to {@code exponent}. */
public long pow(long base, long exponent) { ... }
```

friederbluemle:
I definitely agree with you. I don't like unnecessary repetitions either. However, once you start a comment with `/**` it falls under the Javadoc rules and regulations, which means a `@param` tag for each parameter and a `@return` tag for methods that don't return `void`. The thinking behind this might have been simplified parsing by automated documentation generators and improved visualization of code hints. In any case - more often than not it will cause redundancies when reading through plain source code. If you don't want these in this project, then what is your suggestion? We could try to explicitly suppress the warnings. Based on a quick Google search, it is not straightforward. This might be relevant: [How can I suppress warnings (codebase-wide) during javadoc compilation?](http://stackoverflow.com/questions/620606/how-can-i-suppress-warnings-codebase-wide-during-javadoc-compilation)
I think you agree that a clean build of a freshly cloned project that outputs a bunch of warnings is not ideal.

JakeWharton:
> Javadoc rules and regulations

Where are these defined? If `@param` and `@return` were required doclint would fail with an error, not a warning.

> I think you agree that a clean build of a freshly cloned project that outputs a bunch of warnings is not ideal.

I do not agree that it is a binary choice. If this were Java code we would suppress the warning because while the warning is noted, I am making an explicit choice to ignore it.

I am doing the same here. I value the readability of the Java source file and the Javadoc comment itself over the tool.

friederbluemle:
Sorry for the late reply. These were taken from the Sun/Oracle conventions for writing documentation comments: [How to Write Doc Comments for the Javadoc Tool](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html)
Specifically the section about [Required Tags](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#requiredtags):

> An @param tag is "required" (by convention) for every parameter, even when the description is obvious. The @return tag is required for every method that returns something other than void, even if it is redundant with the method description. (Whenever possible, find something non-redundant (ideally, more specific) to use for the tag comment.)

> These principles expedite automated searches and automated processing. Frequently, too, the effort to avoid redundancy pays off in extra clarity.

Obviously this is different from what you consider good documentation. Feel free to close this PR if you absolutely don't want it :) I created an issue (#265) to keep track of it - and if someone finds an alternate approach to avoid the warnings (by suppressing them?) it can be closed as well.

PR #258, opened 2015-12-18
Fixed custom thread snippet

homer-jay:
The custom thread example fails because there is no message queue attached.

The exception is:
```java
java.lang.RuntimeException: Can't create handler inside thread that has not called Looper.prepare()
```


JakeWharton:
The example should probably use a HandlerThread, but whatever.

PR #269, opened 2016-02-20
BackgroundThreadScheduler - work on thread with lower priority

paveldudka:
As per Android performance guidelines, background work should be scheduled on thread with lower (compared to the main thread) priority, so it doesn't interfere with the main thread. Actually, systrace even has an alert for this:

> Work to produce this frame was descheduled for several milliseconds, contributing to jank. Ensure that code on the UI thread doesn't block on work being done on other threads, and that background threads (doing e.g. network or bitmap loading) are running at android.os.Process#THREAD_PRIORITY_BACKGROUND or lower so they are less likely to interrupt the UI thread. These background threads should show up with a priority number of 130 or higher in the scheduling section under the Kernel process.

BackgroundThreadScheduler is intended to solve this by using custom `ThreadFactory` which creates threads with lowered priority.

You can get an instance of this scheduler by `AndroidSchedulers.backgroundThread()`

### DIFF
--- a/rxandroid/src/main/java/rx/android/plugins/RxAndroidSchedulersHook.java
+++ b/rxandroid/src/main/java/rx/android/plugins/RxAndroidSchedulersHook.java
@@ -35,6 +35,16 @@ public class RxAndroidSchedulersHook {
     }
 
     /**
+     * Scheduler to return from {@link AndroidSchedulers#backgroundThread()} ()} or {@code null} if default
+     * should be used.
+     * <p>
+     * This instance should be or behave like a stateless singleton.
+     */
+    public Scheduler getBackgroundThreadScheduler() {
+        return null;
+    }
+
+    /**
      * Invoked before the Action is handed over to the scheduler.  Can be used for
      * wrapping/decorating/logging. The default is just a passthrough.
      *
--- a/rxandroid/src/main/java/rx/android/schedulers/AndroidSchedulers.java
+++ b/rxandroid/src/main/java/rx/android/schedulers/AndroidSchedulers.java
@@ -18,8 +18,12 @@ import android.os.Looper;
 
 import rx.Scheduler;
 import rx.android.plugins.RxAndroidPlugins;
+import rx.android.schedulers.background.BackgroundThreadScheduler;
+import rx.android.schedulers.handler.HandlerScheduler;
 
-/** Android-specific Schedulers. */
+/**
+ * Android-specific Schedulers.
+ */
 public final class AndroidSchedulers {
     private AndroidSchedulers() {
         throw new AssertionError("No instances");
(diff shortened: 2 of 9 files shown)


JakeWharton:
This can already be done by hooking into RxJava's normal `newThread()` scheduler via the plugins and set a background priority on threads created.

Otherwise, I'm against this:
 * There's nothing Android specific about it which is both a reason against the feature but also a bug: it's not using the correct Android mechanisms for changing thread priority (see your quoted documentation).
 * We should not be encouraging the use of a scheduler which creates a new thread for every worker. It's basically never the right choice.

I'd rather see RxJava augmented to include constructor overloads on `NewThreadScheduler` which allow passing a `ThreadFactory` and a constructor overload on `RxThreadFactory` for supplying a thread priority.

paveldudka:
@JakeWharton By hooking into RxJavaPlugin you mean `RxJavaPlugin#registerSchedulersHook()`? If so, this way you can only override Scheduler returned by `Schedulers.newThread()` call. But you still need to implement this custom scheduler. Or I am missing something?

Regarding other comments:

- Guideline to use threads with lower priority than **Android** main thread in your Android apps, looks pretty Android-related to me. The main reason behind that is RxJava allows you to easily "jump" to the background thread w/o any boilerplate code. And that's what people do - they use `Schedulers.io()`, `Schedulers.newThread()`, `Schedulers.computation()`, etc. That leads to the problem that thread inherits main thread priority and you might hit performance problems. If we have something super simple - people will adopt it really fast. Even in RxAndroid sample app you use HandlerScheduler to jump to the background thread. Just for that you need to extend HandlerThread, override default priority, call start(), get looper out of it and create handler with this looper. That's a lot of boilerplate and honestly, a lot of Android devs don't even know how to do it properly.

Also, not sure using `Thread#setPriority()` API is *incorrect* Android mechanism. This is just another way of changing threads priority (see [official documentation](http://developer.android.com/reference/java/lang/Thread.html#NORM_PRIORITY)).
- Totally agree that we should not be encouraging people to create new thread every time we need to execute something on background, but there are some completely valid use cases for this approach and people are actively using it through RxJava. So if they decide it is OK to spawn a new thread every single time - let's give them an out-of-the-box approach which is optimized for Android

Also I can see the point of augmenting RxJava APIs with possibility to provide priority - probably it will solve the issue on more fundamental level

JakeWharton:
> But you still need to implement this custom scheduler. Or I am missing something?

In my first comment:

> I'd rather see RxJava augmented to include constructor overloads on `NewThreadScheduler` which allow passing a `ThreadFactory` and a constructor overload on `RxThreadFactory` for supplying a thread priority.

The advantage to doing it with the built-in schedulers is that it applies to libraries and all other code you don't control which are already using the schedulers from RxJava itself.

Aside from that,

> So if they decide it is OK to spawn a new thread every single time - let's give them an out-of-the-box approach which is optimized for Android

With the current method naming you're only giving them a rope to hang themselves. At the very least it would need to be named `newBackgroundThread()` to make clear its very, very expensive behavior.

I still can't imagine the case where this would be useful _over_ providing background versions of the pool-backed schedulers like `io()` and `computation()`–perhaps via shipping an implementation of the scheduler hook which configures it for low-priority threads.

paveldudka:
I guess I agree with that. Will go back to my dungeon and come up with something better :)

JakeWharton:
Filed https://github.com/ReactiveX/RxJava/issues/3724

PR #276, opened 2016-04-22
Eliminate allocations by using the Handler to track scheduled actions.

### LINKED ISSUE #275: AndroidScheduler shared vs per-worker Handler
In the current `AndroidScheduler` code, a `Handler` is shared between all workers, and thus each task has to be tracked inside a Worker to remove them upon unsubscription.

I've been playing around with `Handler` for 2.x and I'd think this extra level of tracking could be avoided because the mass unsubscription is available via `handler.removeCallbacksAndMessages(null);`:

```java
static final class MainThreadWorker extends Worker {

    final Handler handler = new Handler(Looper.getMainLooper());

    volatile boolean disposed;

    @Override
    public Disposable schedule(Runnable run, long delay, TimeUnit unit) {
        if (disposed) {
            return EmptyDisposable.INSTANCE;
        }
        if (delay <= 0L) {
            handler.post(run);
        } else {
            handler.postDelayed(run, TimeUnit.MILLISECONDS.convert(delay, unit));
        }
        if (disposed) {
            handler.removeCallbacks(run);
            return EmptyDisposable.INSTANCE;
        }
        return () -> handler.removeCallbacks(run);
    }

    @Override
    public void dispose() {
        disposed = true;
        handler.removeCallbacksAndMessages(null);
     }
}
```

I'm not familiar with Android's performance profile but is there a penalty for having individual Handlers?

JakeWharton:
Closes #275.

cc @akarnokd

### DIFF
--- a/rxandroid/src/main/java/rx/android/schedulers/HandlerScheduler.java
+++ b/rxandroid/src/main/java/rx/android/schedulers/HandlerScheduler.java
@@ -18,11 +18,12 @@ import java.util.concurrent.TimeUnit;
 import rx.Scheduler;
 import rx.Subscription;
 import rx.android.plugins.RxAndroidPlugins;
+import rx.exceptions.OnErrorNotImplementedException;
 import rx.functions.Action0;
-import rx.internal.schedulers.ScheduledAction;
-import rx.subscriptions.CompositeSubscription;
+import rx.plugins.RxJavaPlugins;
 import rx.subscriptions.Subscriptions;
 import android.os.Handler;
+import android.os.Message;
 
 /** A {@link Scheduler} backed by a {@link Handler}. */
 public final class HandlerScheduler extends Scheduler {
@@ -44,10 +45,8 @@ public final class HandlerScheduler extends Scheduler {
     }
 
     static class HandlerWorker extends Worker {
-
-        final Handler handler;
-
-        private final CompositeSubscription compositeSubscription = new CompositeSubscription();
+        private final Handler handler;
+        private volatile boolean unsubscribed;
 
         HandlerWorker(Handler handler) {
             this.handler = handler;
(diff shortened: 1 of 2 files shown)


rxandroid/src/main/java/rx/android/schedulers/HandlerScheduler.java
```diff
@@ -44,45 +45,49 @@ public Worker createWorker() {
...
 
-            handler.postDelayed(scheduledAction, unit.toMillis(delayTime));
+            Message message = Message.obtain(handler, scheduledAction);
+            message.obj = this; // Used as token for unsubscription operation.
 
-            scheduledAction.add(Subscriptions.create(new Action0() {
-                @Override
-                public void call() {
-                    handler.removeCallbacks(scheduledAction);
+            synchronized (this) {
```
akarnokd: Does this have to be synchronized? Why not do the post-check after the submission and remove token+message pair?

rxandroid/src/main/java/rx/android/schedulers/HandlerScheduler.java
```diff
@@ -44,45 +45,49 @@ public Worker createWorker() {
...
     static class HandlerWorker extends Worker {
-
-        final Handler handler;
-
-        private final CompositeSubscription compositeSubscription = new CompositeSubscription();
+        private final Handler handler;
+        /* Writes are guarded by 'this'. */
+        private volatile boolean unsubscribed;
 
         HandlerWorker(Handler handler) {
```
artem-zinnatullin: Please correct me if I'm wrong but the idea in #275 was to create `Handler` per `Worker` to perform mass cancellation of the tasks. At the moment, you share one `Handler` between all `HandlerWorkers` and `unsubscribe()` will cancel tasks of other `HandlerWorker`s too which is incorrect behavior.

JakeWharton: A token is used to cancel.

artem-zinnatullin: oh nice, now I see 👍 

// stuck to idea that `Handler` per `Worker` is required, your approach is better

rxandroid/src/main/java/rx/android/schedulers/HandlerScheduler.java
JakeWharton: Nope, that'll work too. Updated.

rxandroid/src/main/java/rx/android/schedulers/HandlerScheduler.java
JakeWharton: I'll add a test to ensure multiple workers don't interfere with each other though.

artem-zinnatullin: Just wanted to ask for that, thx

akarnokd:
:+1: looks good to me

rxandroid/src/main/java/rx/android/schedulers/HandlerScheduler.java
JakeWharton: added

rxandroid/src/test/java/rx/android/schedulers/HandlerSchedulerTest.java
```diff
@@ -221,4 +220,24 @@ public void call() {
 
         assertTrue(neverCalled.get());
     }
+
+    @Test
+    public void schedulerWorkerCancellationDoesNotAffectOtherWorkers() {
```
artem-zinnatullin: 👍

PR #277, opened 2016-04-22
Rewrite HandlerScheduler tests to be black box.

JakeWharton:
Previously we mocked a Handler and relied on verifying the correct methods were called. This meant that changing the implementation of HandlerScheduler meant changing the tests. This is prone to masking errors since it focuses on *what* the implementation does instead of *how* it behaves. All we care about is behavior, and all these new tests do is verify behavior.

It should be noted that these new tests (just like the old ones) are still reliant on the behavior of Robolectric's fake Looper.


JakeWharton:
Also added some tests to increase logical coverage.

dlew:
LGTM.

hzsweers:
👍

PR #280, opened 2016-05-01
Cache the hook reference upon worker creation.

JakeWharton:
There's no need to look it up each time since it can't change.


dlew:
What if someone calls `RxAndroidPlugins.reset()`? Or is that strictly for testing purposes?

JakeWharton:
Then they wait for the next worker to be created in order for those changes
to be reflected. I don't think it's worth optimizing for that specific case
because I do think it's only for tests. The RxJava schedulers are caching
the hook in the worker as well.

dlew:
I agree. 👍  from me.

PR #282, opened 2016-05-01
Change scheduler factory to care about loopers, not handlers.

JakeWharton:
A Looper is the currency of Android-specific work scheduling that we should be exposing. Handler is only a helper through which we access the Looper.


rxandroid/src/main/java/rx/android/schedulers/HandlerScheduler.java
```diff
@@ -13,122 +13,29 @@
...
 
-/** A {@link Scheduler} backed by a {@link Handler}. */
-public final class HandlerScheduler extends Scheduler {
-    /** Create a {@link Scheduler} which uses {@code handler} to execute actions. */
+/**
+ * A {@link Scheduler} backed by a {@link Handler}.
+ *
+ * @deprecated Use {@link AndroidSchedulers#from(Looper)}.
+ */
+@Deprecated
```
dlew: If we're going to another major revision for the next release (2.x), why not just remove it entirely?

dlew:
LGTM besides the question of whether to deprecate or remove.

rxandroid/src/main/java/rx/android/schedulers/HandlerScheduler.java
JakeWharton: Next is 1.2. I don't see a reason to do a major release for it and having
major versions out of sync with RxJava is going to be a mess.

JakeWharton:
I filed the 2.x issues for the hypothetical future of what needs done for RxJava 2.0–should it ever happen.

dlew:
Ah, okay, makes sense. In that case deprecation is good.

PR #287, opened 2016-05-02
Fix AndroidSchedulers to create an instance using hook only once.

JakeWharton:
Previously we were asking the hook for an instance every time `mainThread()` was called. This not only impacted performance, but also broke the contract the that hook was a factory–instead requiring that it behave like a thread-safe, lazily-initialized instance cache.

If for whatever reason you do need to change the instance over time, return a scheduler instance which delegates to another and allows swapping out the delegate.


dlew:
👍

hzsweers:
:+1:

PR #292, opened 2016-05-10
Update to RxJava 1.1.5 to get filter() bugfix

EddieRingle:
RE: ReactiveX/RxJava#3912

Signed-off-by: Eddie Ringle <eddie@ringle.io>


JakeWharton:
We never make releases for RxJava bumps, FYI.

PR #313, opened 2016-08-05
Latest RxJava.

JakeWharton:


JakeWharton:
I feel obligated to tell myself that I won't be making a release to track this version.

PR #314, opened 2016-08-05
Latest Robolectric.

JakeWharton:


vanniktech:
@JakeWharton there's actually 3.1.2

PR #315, opened 2016-08-05
Latest Robolectric.

JakeWharton:


artem-zinnatullin:
Just wanted to note that such PR titles and commit messages are not helpful when you need to find something in GitHub/git history…

JakeWharton:
You will never be searching for when Robolectric changed from 3.1.1 to
3.1.2.

artem-zinnatullin:
Ok

PR #318, opened 2016-08-09
Remove OnErrorNotImplementedException handling, track interface rename.

JakeWharton:


JakeWharton:
Merging since nothing controversial here and I want to see a snapshot published.

Issue #346, opened 2016-11-05
IllegalArgumentException: delay < 0: -237000000


PaulWoitaschek:
I received a crash after updating to RxJava 2.0.0

```
java.lang.IllegalStateException: Fatal Exception thrown on Scheduler.
	at io.reactivex.android.schedulers.HandlerScheduler$ScheduledRunnable.run(HandlerScheduler.java:113)
	at android.os.Handler.handleCallback(Handler.java:815)
	at android.os.Handler.dispatchMessage(Handler.java:104)
	at android.os.Looper.loop(Looper.java:194)
	at android.app.ActivityThread.main(ActivityThread.java:5637)
	at java.lang.reflect.Method.invoke(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:372)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:960)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:755)
Caused by: java.lang.IllegalArgumentException: delay < 0: -237000000
	at io.reactivex.android.schedulers.HandlerScheduler$HandlerWorker.schedule(HandlerScheduler.java:60)
	at io.reactivex.Scheduler$Worker$PeriodicTask.run(Scheduler.java:307)
	at io.reactivex.android.schedulers.HandlerScheduler$ScheduledRunnable.run(HandlerScheduler.java:111)
```

samueltardieu:
I have the same issue when using `Flowable.interval` on `AndroidSchedulers.mainThread()`, or `schedulePeriodically` on this scheduler. I do not have the problem when using `Schedulers.computation()` for example.

See https://github.com/cgeo/cgeo/issues/6087 for an example of this bug.

PaulWoitaschek:
I'm receiving many many crashes of this type and bad user reviews. Can I do anything to speed up the fixing of this?

JakeWharton:
Create a failing test case with a fix and send a PR.

akarnokd:
Looks like your periodic task runs too long and the next round should have happened 237 milliseconds ago.

Otherwise, the `Scheduler` spec states that negative values should be considered non-delayed runs:
https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/Scheduler.java#L196
https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/Scheduler.java#L215

[...]

PR #347, opened 2016-11-06
Schedule tasks with negative delays immediately as per Scheduler spec

samueltardieu:
This fixes the issue in https://github.com/ReactiveX/RxAndroid/issues/346.

### DIFF
--- a/rxandroid/src/test/java/io/reactivex/android/schedulers/HandlerSchedulerTest.java
+++ b/rxandroid/src/test/java/io/reactivex/android/schedulers/HandlerSchedulerTest.java
@@ -73,6 +73,15 @@ public final class HandlerSchedulerTest {
     }
 
     @Test
+    public void directScheduleOnceWithNegativeDelayPostsImmediately() {
+        CountingRunnable counter = new CountingRunnable();
+        scheduler.scheduleDirect(counter, -1, TimeUnit.MINUTES);
+
+        runUiThreadTasks();
+        assertEquals(1, counter.get());
+    }
+
+    @Test
     public void directScheduleOnceUsesHook() {
         final CountingRunnable newCounter = new CountingRunnable();
         final AtomicReference<Runnable> runnableRef = new AtomicReference<>();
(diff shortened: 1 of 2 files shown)


rxandroid/src/test/java/io/reactivex/android/schedulers/HandlerSchedulerTest.java
```diff
@@ -688,18 +702,6 @@ public void directSchedulePeriodicallyInputValidation() {
...
-            scheduler.schedulePeriodicallyDirect(new CountingRunnable(), -1, 1, MINUTES);
-            fail();
-        } catch (IllegalArgumentException e) {
-            assertEquals("delay < 0: -1", e.getMessage());
-        }
-        try {
-            scheduler.schedulePeriodicallyDirect(new CountingRunnable(), 1, -1, MINUTES);
-            fail();
-        } catch (IllegalArgumentException e) {
-            assertEquals("period < 0: -1", e.getMessage());
```
JakeWharton, line 700: This should not be removed. A negative period makes no sense.

rxandroid/src/test/java/io/reactivex/android/schedulers/HandlerSchedulerTest.java
```diff
@@ -746,18 +742,6 @@ public void workerSchedulePeriodicallyInputValidation() {
...
-            worker.schedulePeriodically(new CountingRunnable(), -1, 1, MINUTES);
-            fail();
-        } catch (IllegalArgumentException e) {
-            assertEquals("delay < 0: -1", e.getMessage());
-        }
-        try {
-            worker.schedulePeriodically(new CountingRunnable(), 1, -1, MINUTES);
-            fail();
-        } catch (IllegalArgumentException e) {
-            assertEquals("period < 0: -1", e.getMessage());
```
JakeWharton, line 758: This should not be removed. A negative period makes no sense.

JakeWharton (approved):

rxandroid/src/test/java/io/reactivex/android/schedulers/HandlerSchedulerTest.java
samueltardieu: Where is this error message supposed to come from? Grepping RxAndroid and RxJava for "period < 0" gives no result. Moreover, wouldn't such test belong in RxJava?

I thought that these tests were disabled anyway, but I can reinstate them if you prefer.

JakeWharton: The test is ignored because it's currently not implemented natively. That still doesn't change the fact that its removal is unrelated to this PR.

samueltardieu: Fair enough. I will amend the PR.

samueltardieu: @JakeWharton This has been amended, I don't know if you expect a ping, but in any case here it is.

JakeWharton:
I'll do a point release tomorrow or Friday. Thanks!

PR #373, opened 2017-04-21
Update README.md

manijshrestha:
update version in documentation


JakeWharton:
2.0.9 comes out tomorrow. But we're at least up-to-date for 24 hours!

akarnokd:
Just 1 more week and 2.1 is out. It would be great RxAndroid also had a release with updated dependency.

JakeWharton:
Yep! That seems reasonable.

PR #406, opened 2017-11-27
Update RxJava2 version in README.md

nakker1218:
Update RxJava version. `2.1.5 -> 2.1.7`


artem-zinnatullin:
Tip for future: pls don't use such generic commit messages/PR titles, it makes git history much harder to analyze, thanks :)

PR #408, opened 2018-01-03
Remove superfluous negative check.

JakeWharton:
Handler does this check already since 2009 when AOSP was imported to git (aka forever).


akarnokd (approved):

artem-zinnatullin (approved):
Nice

vanniktech (approved):

PR #415, opened 2018-02-12
Disable BuildConfig generation.

JakeWharton:


artem-zinnatullin (approved):
Cloned, built, confirmed that aar doesn't contain BuildConfig.

:+1:

tebasakyu:
👀

PR #416, opened 2018-02-12
Proposal/rehash: Use async Messaging

ZacSweers:
This PR changes the default main thread `Handler` to use asynchronous messaging. This allows us to avoid VSYNC locking that otherwise pushes every `post` to the next frame.

This works by relying on the new `Handler.createAsync` factory in API 28, and on pre-28 it will reflectively fall back to a private constructor of Handler (that's always existed) to enable this.

This should give us the original goal of #228 with the safety of still leaning on the main thread looper and not trying to directly run things ourselves and risking deadlocks/trampolining issues.

To use SDK 28, updated the AGP plugin and gradle along the way.

*Prior discussion below*

~This is a main thread scheduler that's optimized to directly execute given Runnables if it's already on the main thread, rather than always scheduling.~

~Prior art: #228~

~Kept as a standalone implementation right now for review and discussion, but ideally this implementation would replace the existing main thread scheduler implementation as the default.~


ZacSweers:
Note that this is upstreaming from our internal implementation at Uber. We've been using this in our production apps for a little over a year now.

From some discussion with @JakeWharton offline: does it make sense for a scheduler that doesn't always necessarily "schedule"? Some potential precedent of this in RxJava include `TrampolineScheduler`'s implementation and `ImmediateThinScheduler`. @akarnokd would really love to get your thoughts on this!

akarnokd:
I think I expressed my views in the prior art. Doing it plainly may result in a stack overflow, adding a main-thread trampoline may add overhead negating the optimization and supporting task disposal may become quite complicated.

ZacSweers:
Agreed the main thread trampoline adds a lot of overhead. Is it really necessary though? It seems like that risks duplicating what the main thread looper does under the hood.

For stack overflow, is that due to stacktraces no longer being chopped? I didn't quite follow that bit

akarnokd:
```java
Worker w = new FastPathScheduler().createWorker();

Schedulers.single().scheduleDirect(() -> w.dispose(), 10, TimeUnit.SECONDS);

w.schedule(new Runnable() {
    @Override public void run() {
        w.schedule(this);
    } 
});

Thread.sleep(11000);
```

Trampoline would limit the stack depth.

damianw:
Doesn't this potentially violate the serial requirement for schedulers? If you have an upstream observable that emits a value which must be scheduled on the handler, followed closely by an emission  which can be immediate (or trampolined), the second emission could easily pass through to the subscriber before the first.

My first thought for solving this problem would be to only take the fast path when there are no queued emissions.

ZacSweers:
Hey all - after a lot of discussion offline and with the Android framework team, I've updated the PR with a new solution that relies on asynchronous `Messages` to avoid vsync locking while keeping the safety of the main thread looper. I've updated the description as well, please take a look!

ZacSweers:
I hate Travis CI sometimes. Will debug more after allowing time for discussion

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
```diff
@@ -47,6 +48,28 @@ public static Scheduler from(Looper looper) {
         return new HandlerScheduler(new Handler(looper));
     }
 
+    @TargetApi(Build.VERSION_CODES.P)
```
JakeWharton: This shouldn't be needed

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
```diff
@@ -47,6 +48,28 @@ public static Scheduler from(Looper looper) {
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            return Handler.createAsync(looper);
+        } else {
+            try {
+                //noinspection JavaReflectionMemberAccess
+                return Handler.class
+                    .getConstructor(Looper.class, Handler.Callback.class, boolean.class)
+                    .newInstance(looper, null, true);
+            } catch (IllegalAccessException e) {
+                return new Handler(looper);
```
JakeWharton: Just make these all empty and hoist a single `return` outside.

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
```diff
@@ -47,6 +48,28 @@ public static Scheduler from(Looper looper) {
         return new HandlerScheduler(new Handler(looper));
     }
 
+    @TargetApi(Build.VERSION_CODES.P)
+    private static Handler createAsyncHandler(Looper looper) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            return Handler.createAsync(looper);
+        } else {
```
JakeWharton: `else` not required after `return`

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
```diff
@@ -13,20 +13,21 @@
...
+import java.util.concurrent.Callable;
 
 /** Android-specific Schedulers. */
 public final class AndroidSchedulers {
 
     private static final class MainHolder {
-
-        static final Scheduler DEFAULT = new HandlerScheduler(new Handler(Looper.getMainLooper()));
+        static final Scheduler DEFAULT
+            = new HandlerScheduler(createAsyncHandler(Looper.getMainLooper()));
```
JakeWharton: This seems like it's going to require a major version update. The implications to the entirety of the world which uses RxAndroid are scary and a silent change in a minor release doesn't seem appropriate.

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
ZacSweers: I agree! But lint fails without it despite the if-check :|

JakeWharton: Did you try just using `28` as a literal? I prefer avoiding the codenames (which are thankfully almost dead) and letters (which we should be trying to kill next).

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
ZacSweers: Good catch 551b7b4

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
ZacSweers: 2d46b26

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
ZacSweers: Huh, yeah using just 28 worked. c9e2bbf

RIP dessert club

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
ZacSweers: I agree, but a major change would require breaking with keeping the same major version as rxjava :|.

Either that or maybe do an extended preview release and ask for feedback?

JakeWharton: I think we overload `from(Looper)` with a version that takes a boolean for `async` just like the `Handler` constructor no matter what.

As to getting it into `mainThread()`, you can use the `RxAndroidPlugins` to install an init callback for its creation which you can then delegate to `from(Looper.getMainLooper(), true)`. 

I think that's the minimum-viable API to at least get this into a release now for people to start testing it. And from there we can determine whether or not it's even possible to make it the default behind `mainThread()` in the next major version–or sometime in the future–and whether or not we need to expose a `vsyncMainThread()` or similar accessor to a singleton instance for code which truly needs the old behavior.

ZacSweers: That works too! For some reason I had it in my head that you were against having support for both. Can update. Should I add a brief mention/example in the readme with it or save that for changelog?

JakeWharton:
Can we get rid of the use of `Handler` and just send raw `Message` instances to the `Looper`? Then we can use `Message.setAsynchronous(boolean)` which doesn't require reflection?

ZacSweers:
We could but that API was only added (made public?) on API 22. If it was just private before we'd still need reflection.

JakeWharton:
The functionality didn't exist before 22, just like the constructor on Handler you're currently reflecting into:

 * https://android.googlesource.com/platform/frameworks/base/+/e799cb78b4be61d3882e71c6812fa62c9a83fd5d
* https://android.googlesource.com/platform/frameworks/base/+/a2910d0abbbe18ba1710dfd4a31af45769632255

ZacSweers:
Oh wtf. I could have sworn I went back and saw the handler constructor on older APIs (my understanding was it existed with the introduction of project butter circa API 17). In that case, yeah we can avoid reflection. Probably doesn't make sense to use the new `createAsync` API on 28+ either to keep things simple, but it'd almost certainly be a breaking change or a new separate scheduler that has the `Message`-based API. How would you want that to look?

JakeWharton:
We already use `Message`. Looks like all you need to do is keep a boolean in `HandlerScheduler` and set it here:

https://github.com/ReactiveX/RxAndroid/blob/5cb239130dceab95797b7313b19f2d8e95407d8f/rxandroid/src/main/java/io/reactivex/android/schedulers/HandlerScheduler.java#L69-L72

ZacSweers:
So keep the handler then? I was thrown off by

> Can we get rid of the use of Handler

ZacSweers:
We'd also have to promote `HandlerScheduler` to be a public API if you want it to be optional and usable via plugin, which it isn't right now

JakeWharton:
The boolean comes in through `from` which creates the `HandlerScheduler`.
That's the only public API change.

adamp:
Hi, chiming in from the Android team to give a general thumbs up to skipping the vsync barriers here. Ordering between "async" messages is still preserved; we made waiting until after the barriers the default behavior back in Jellybean because there's a bunch of UI code out there that expects a message or runnable to be processed after any pending view tree traversals (measure/layout/draw) which wait for vsync after Jellybean. The tradeoff for the compatibility is a lot of latency that most use cases shouldn't have to pay for, which this proposal addresses.

The Message.setAsynchronous method has been in the codebase since API 16 (jellybean - https://android.googlesource.com/platform/frameworks/base/+/jb-release/core/java/android/os/Message.java#397) and Handler's constructor with the boolean async param has been there since API 17 (jellybean-mr1 - https://android.googlesource.com/platform/frameworks/base/+/jb-mr1-release/core/java/android/os/Handler.java). It would be safe to opportunistically call either one of these using reflection (or for Message.setAsynchronous just by building against a newer sdk and using a different API level check/catch if something goes wrong) as of these API levels, provided that when you're running on a platform version where it's officially available you call the real public API. This is the policy we use for reflection in the Android support libraries - as long as the implementation is public API clean as of some published API level the use of reflection isn't defining de facto public API that we have to support into the future.

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
```diff
@@ -43,8 +44,31 @@ public static Scheduler mainThread() {
...
+        Handler handler = new Handler(looper);
+        if (useAsync) {
+            Message message = Message.obtain(handler);
+            try {
+                message.setAsynchronous(true);
+            } catch (NoSuchMethodError e) {
+                useAsync = false;
+            }
+        }
+        return new HandlerScheduler(new Handler(looper), useAsync);
```
JakeWharton: You already created the `Handler` above. Can pass it here.

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
```diff
@@ -43,8 +44,31 @@ public static Scheduler mainThread() {
...
+        boolean useAsync = async && Build.VERSION.SDK_INT >= 16;
+        // Confirm the method is there
+        Handler handler = new Handler(looper);
+        if (useAsync) {
+            Message message = Message.obtain(handler);
+            try {
+                message.setAsynchronous(true);
+            } catch (NoSuchMethodError e) {
+                useAsync = false;
+            }
```
JakeWharton: We should recycle the `Message` to be a good citizen!

JakeWharton (approved):

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
ZacSweers: d9a51a7

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
ZacSweers: derp, 6fbdff3

ZacSweers:
Alright, reverted the sdk 28 changes to punt the CI issues for now and also tweaked it to only do the `setAsynchronous` method check on API < 22. This shooooould be good to go now

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
```diff
@@ -43,8 +44,33 @@ public static Scheduler mainThread() {
...
+    }
+
+    /**
+     * A {@link Scheduler} which executes actions on {@code looper}.
+     *
+     * @param async if true, the scheduler will use async messaging on API >= 16 to avoid VSYNC
+     *              locking. On API < 16, this will no-op.
+     * @see android.os.Message#setAsynchronous(boolean)
+     */
+    @SuppressLint("NewApi")
```
JakeWharton: Is this still needed?

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
```diff
@@ -43,8 +44,33 @@ public static Scheduler mainThread() {
...
     public static Scheduler from(Looper looper) {
+        return from(looper, false);
+    }
+
+    /**
+     * A {@link Scheduler} which executes actions on {@code looper}.
+     *
+     * @param async if true, the scheduler will use async messaging on API >= 16 to avoid VSYNC
+     *              locking. On API < 16, this will no-op.
+     * @see android.os.Message#setAsynchronous(boolean)
```
JakeWharton: shouldn't need to specify package since this is now imported

JakeWharton (approved):

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
JakeWharton: I guess because the conditional is reversed it is

ZacSweers: Yeah :/

rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
ZacSweers: 4ef9c53

JakeWharton:
Merged as 191a758fcce4e4d16ad0438ad4dac3fd01b4362f

ahulyk:
to try that feature i need to init it in the Application onCreate() method, right?
```
AndroidPlugins.setInitMainThreadSchedulerHandler {
    AndroidSchedulers.from(Looper.getMainLooper(), true)
}
```

limpep:
@ahulyk here is an article on how to use it, https://medium.com/@sweers/rxandroids-new-async-api-4ab5b3ad3e93

Issue #424, opened 2018-05-02
It is possible for HandlerScheduler$ScheduledRunnable.run to still be called on a disposed object


stevelilly:
I began investigating this as we have been seeing some strange crashes, where seemingly the only explanation was our subscriber was invoked after it was disposed (both the observe and the dispose in the main thread).

I have posted an example project here.
https://github.com/stevelilly/rxandroid-racecondition
Tested against rxandroid:2.0.2 + rxjava:2.1.13

This report is based on the belief that when running the following code in the main thread it should
not be possible to reach the exception: 

```java
Single.fromCallable(Object::new)
    .subscribeOn(Schedulers.io())
    .observeOn(AndroidSchedulers.mainThread())
    .subscribe(object -> {
        throw new IllegalStateException("should never be called");
    })
    .dispose();
```

After a few thousand calls I often encounter the exception. More interestingly, in the debugger
we can see that the `HandlerScheduler$ScheduledRunnable` has already been disposed, yet its `run()` method is still called.

![Debugging run() on disposed Runnable](https://github.com/stevelilly/rxandroid-racecondition/blob/master/runOnDisposedRunnable.png?raw=true)

AFAICT the code in the `HandlerScheduler` is using `removeCallbacks` and `removeCallbacksAndMessages` in the proper way, however it appears it is not safe to call `delegate.run()` without first checking `disposed`.

This seems to affect API 24+ much more often than API 23 when I test it, but this could just be the
specific timings of my emulator.

JakeWharton:
It's not clear to me how `disposed` can be set to `true` from a call to `dispose()` on the main thread and the `Runnable` **not** be removed from the message queue of the main looper. We shouldn't need the proposed check since the main thread guarantees the ordering of these operations. Were this a background disposal the behavior would make sense since it becomes a race at that point (and we let the race happen inside the message queue rather than re-checking the boolean ourselves).

artem-zinnatullin:
Looks like it happens due to [call to `DisposableHelper.replace()` in `SingleObserveOn`](https://github.com/ReactiveX/RxJava/blob/v2.1.13/src/main/java/io/reactivex/internal/operators/single/SingleObserveOn.java#L65).

Because of that call, call to `HandlerScheduler.ScheduledRunnable.dispose()` can actually end up on io thread (in this particular example), thus race with main thread that could already start executing the runnable.

A fix could be implemented either in RxAndroid (check `disposed` in `run()` as suggested) or in RxJava. 
[...]

PR #425, opened 2018-05-03
Remove handler callbacks before setting disposed.

JakeWharton:
This change ensures that you can never observe `disposed` being set to `true` inside `run()`. `removeCallbacks` and the mechanism by which a `Looper` retrieves the head of the message queue are both governed by a lock. When `dispose()` is called on a non-main thread taking this lock is the race which determines cancelation and **not** the boolean.

Include a comment emphasizing that the boolean is tracked solely for the purposes of accurate `isDisposed()` reporting.

Refs #424.

### DIFF
--- a/rxandroid/src/main/java/io/reactivex/android/schedulers/HandlerScheduler.java
+++ b/rxandroid/src/main/java/io/reactivex/android/schedulers/HandlerScheduler.java
@@ -96,7 +96,7 @@ final class HandlerScheduler extends Scheduler {
         private final Handler handler;
         private final Runnable delegate;
 
-        private volatile boolean disposed;
+        private volatile boolean disposed; // Tracked solely for isDisposed().
 
         ScheduledRunnable(Handler handler, Runnable delegate) {
             this.handler = handler;
@@ -114,8 +114,8 @@ final class HandlerScheduler extends Scheduler {
 
         @Override
         public void dispose() {
-            disposed = true;
             handler.removeCallbacks(this);
+            disposed = true;
         }
 
         @Override


JakeWharton:
We could optionally make the `Handler` reference nullable and volatile and set it to `null` to save the `boolean`.

artem-zinnatullin:
Hm, I'm not sure that reordering is a good idea, `Handler.removeCallback` takes much more time than writing to a volatile field. I think we want to make `disposed` change visible asap. It's not harmful though, as `Handler.removeCallback` is going to noop on subsequent calls. 

Also I'm pretty sure it's still possible to see `disposed == true` in `run()` using code from referenced issue. 
There is a race between executing `ScheduledRunnable` on main thread and `dispose()` from io thread.

I have an idea on how to not do `dispose` on upstream thread which should remove the race in first place for referenced issue, but it requires an additional field and I haven't tested it yet, so I might be wrong.

// I'm on mobile, can elaborate later

JakeWharton:
It takes time because it's the thing that's actual doing the cancelation. The boolean is redundant information solely for internal tracking purposes.

JakeWharton:
Moreover, setting the boolean before `removeCallbacks` completes can be argued as a bug as you can observe `isDisposed()` true on another thread before the lock on the message queue is taken.

JakeWharton:
Ah, I realize what you're likely going to post as I don't think I was clear. It's _always_ the case that the boolean can be observed to be set while the runnable is running when they run concurrently. What this change does, however, is eliminate the ability for the boolean to reflect a state that's different than the _real_ underlying race. That is, either `Handler.removeCallbacks` or `MessageQueue.next` is going to win and that becomes the decision to run or not. If you are disposing concurrently there is _always_ a race. If you are disposing on the same looper as the handler then you have no problems. The fact that RxJava causes a race where you otherwise don't expect one is not this library's problem as it affects all schedulers.

ZacSweers (approved):

PR #429, opened 2018-06-08
Replace null check with requireNonNull

shenghaiyang:


JakeWharton:
This function is not part of RxJava's public API and so we cannot rely on it being present.

PR #442, opened 2018-09-05
considers async option in HandlerScheduler.scheduleDirect

### LINKED ISSUE #441: HandlerScheduler.scheduleDirect supports async option?
RxAndroid support async option v2.1.0. now, it support only in `HandlerWorker.schedule`.

I think, `HandlerScheduler.scheduleDirect` should be support async option.
Because, some RxJava operator use `Scheduler.scheduleDirect` method (ex, Maybe.observeOn)

(i prepared source code in https://github.com/ReactiveX/RxAndroid/compare/2.x...satoshun:2.x)

satoshun:
close https://github.com/ReactiveX/RxAndroid/issues/441

### DIFF
--- a/rxandroid/src/main/java/io/reactivex/android/schedulers/HandlerScheduler.java
+++ b/rxandroid/src/main/java/io/reactivex/android/schedulers/HandlerScheduler.java
@@ -32,13 +32,18 @@ final class HandlerScheduler extends Scheduler {
     }
 
     @Override
+    @SuppressLint("NewApi") // Async will only be true when the API is available to call.
     public Disposable scheduleDirect(Runnable run, long delay, TimeUnit unit) {
         if (run == null) throw new NullPointerException("run == null");
         if (unit == null) throw new NullPointerException("unit == null");
 
         run = RxJavaPlugins.onSchedule(run);
         ScheduledRunnable scheduled = new ScheduledRunnable(handler, run);
-        handler.postDelayed(scheduled, unit.toMillis(delay));
+        Message message = Message.obtain(handler, scheduled);
+        if (async) {
+            message.setAsynchronous(true);
+        }
+        handler.sendMessageDelayed(message, unit.toMillis(delay));
         return scheduled;
     }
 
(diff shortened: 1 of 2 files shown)


JakeWharton:
This needs tests. Are you able to write them? If not, I can get to it sometime in the next week or two.

satoshun:
oh sorry, i forget writing test...

But, I think already [HandlerSchedulerTest](https://github.com/ReactiveX/RxAndroid/blob/2.x/rxandroid/src/test/java/io/reactivex/android/schedulers/HandlerSchedulerTest.java#L55) is considering the async option test by Parameterized test.

Are there other missing tests? Please tell me if any missing test.

JakeWharton:
Coverage is definitely missing. Either the test should have been failing before without this change or we aren't covering this codepath. Since the test is passing, we have missing coverage on this codepath.

ZacSweers:
The parameterized test just expects no regressions. There weren't any cases added that would have failed the the non-async implementation, as I'm not really sure how to coerce vsync behavior out of robolectric :X

JakeWharton:
You don't need to. We just need to assert the flag is set through all scheduler code paths. There's already a test which asserts that flag is not set on pre-16 which I added.

ZacSweers:
Ah didn't know that was added 😬

JakeWharton:
We just need to add ones which check each scheduler method and assert the flag is set to true on the messages. We should have been doing that anyway...

satoshun:
thx 😄 

> assert the flag is set to true on the messages.

i write a test!

satoshun:
@JakeWharton @hzsweers i added test for async option 😃

ZacSweers (approved):
Lgtm. Not terribly familiar with those robolectric APIs but not sure of any better ways anyway

ZacSweers:
Thanks!

jj-crypto (approved):

AaronWashington (commented):
[io.reactivex.rxjava3:rxandroid:3.0.0](io.reactivex.rxjava3:rxandroid:3.0.0)

PR #534, opened 2019-06-20
Update dependencies and Java 8 support

lsanotes:


build.gradle
```diff
@@ -21,7 +21,7 @@ subprojects {
 }
 
 ext {
-    minSdkVersion = 9
+    minSdkVersion = 14
     compileSdkVersion = 28
-    sourceCompatibility = JavaVersion.VERSION_1_7
+    sourceCompatibility = JavaVersion.VERSION_1_8
 }
```
JakeWharton, line 27: I don't want to change this. The library is effectively done and won't benefit from any language feature of 8 in a meaningful way.

build.gradle
```diff
@@ -21,7 +21,7 @@ subprojects {
...
 
 ext {
-    minSdkVersion = 9
+    minSdkVersion = 14
```
JakeWharton, line 24: Similarly, there doesn't seem to be a reason to change this

lsanotes: This change is caused by upgrading robolectric to 4.3. It integrated the androidx.test framework, which requires min sdk 14.

lsanotes: http://robolectric.org/migrating/

lsanotes:
I'll split the pull request to smaller ones.

PR #540, opened 2019-09-02
Update to RxJava 3

### LINKED ISSUE #538: RxJava 3 support
How to use this with RxJava 3??
I'm getting duplicate files error on build.
Thank you.

drewhamilton:
Closes #538.

With RxJava 3.0.0-RC2, the package names of all RxJava types were changed. This PR updates all the imports in RxAndroid accordingly.

If the intention is to update RxAndroid's package names as well I'm happy to add that to this PR.

This shouldn't actually be merged to 2.x, there just isn't a 3.x to point to yet.

### DIFF
--- a/rxandroid/build.gradle
+++ b/rxandroid/build.gradle
@@ -36,7 +36,7 @@ android {
 }
 
 dependencies {
-    api 'io.reactivex.rxjava2:rxjava:2.2.10'
+    api 'io.reactivex.rxjava3:rxjava:3.0.0-RC2'
 
     testImplementation 'junit:junit:4.12'
     testImplementation 'org.robolectric:robolectric:3.8'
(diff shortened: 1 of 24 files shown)


vanniktech (commented):
You'd also want to change the package names of this library so RxAndroid 2.x and 3.x can live side by side

JakeWharton:
Please retarget to 3.x branch

drewhamilton:
@vanniktech @JakeWharton `io.reactive.rxjava3.android.**`?

Also: Should `MainThreadDisposable` be moved to a `disposables` package to align with RxJava's packaging conventions?

JakeWharton:
Since we offer dramatically less API surface, we don't need the same package layering.

That package seems like the right one to me.

PR #544, opened 2019-09-22
Comment for posterity why androidx-core isn't used for async

ZacSweers:
From personal experience, this kind of detail pointing to something Official™️ can save developers a lot of headache having to explain to skeptical colleagues.


JakeWharton:
Weird casing of "Androidx-core"

ZacSweers:
Didn’t notice that. Originally I wrote just Androids then specified core and forgot to lower. Can fix

JakeWharton:
I wouldn't have merged if I cared enough to see it changed

ZacSweers:
Too late! #546

PR #338, opened 2016-10-05
2.x: Add lazy loading of default main Scheduler

peter-tackage:
Initial proposal to address #332 

I've essentially used the same implementation as in RxJava: https://github.com/ReactiveX/RxJava/pull/4585.

For the sake of common implementation, I've used RxJava's internal(?) classes `ObjectHelper` and `ExceptionHelper`. Happy to consider alternatives if we don't want this.

PR #571, opened 2020-06-11
Remove unnecessary holder for the singleton of MainScheduler

yhuangh:
The static field MAIN_THREAD will be initialized first when class AndroidSchedulers load. So field of Holder will be used at the same. There's no need to wrap it with a Holder class.


JakeWharton:
Please see https://github.com/ReactiveX/RxAndroid/pull/338/files#r82259113 where I had the very same thought when it was added. At this point, I would prefer not to change it simply for stability purposes. At this time there's no plans to have another release of the library anyway.

PR #576, opened 2020-12-23
Update gradle

Goooler:
- update dependencies
- update root `build.gradle`

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,6 @@
 android.enableJetifier=false
 android.useAndroidX=true
 org.gradle.jvmargs=-Xmx2048m
-org.gradle.parallel=true
 
 GROUP=io.reactivex.rxjava3
 VERSION_NAME=3.0.1
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,5 @@
-android.enableJetifier=true
-android.injected.testOnly=false
+android.enableJetifier=false
 android.useAndroidX=true
-org.gradle.caching=true
 org.gradle.jvmargs=-Xmx2048m
 org.gradle.parallel=true
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,10 @@
+android.enableJetifier=true
+android.injected.testOnly=false
+android.useAndroidX=true
+org.gradle.caching=true
+org.gradle.jvmargs=-Xmx2048m
+org.gradle.parallel=true
+
 GROUP=io.reactivex.rxjava3
 VERSION_NAME=3.0.1
 POM_NAME=RxAndroid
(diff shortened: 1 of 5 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-all.zip
```
JakeWharton:
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-bin.zip
```

gradle.properties
```diff
@@ -1,3 +1,10 @@
+android.enableJetifier=true
```
JakeWharton: Why is this set to true?

gradle.properties
```diff
@@ -1,3 +1,10 @@
...
+android.injected.testOnly=false
```
JakeWharton: What is this for?

gradle.properties
```diff
@@ -1,3 +1,10 @@
...
+android.useAndroidX=true
+org.gradle.caching=true
```
JakeWharton: I would prefer this to be off. This projects builds in a few seconds so caching doesn't really help. And Gradle doesn't have the best track record for correctness in caching.

JakeWharton (approved):

gradle.properties
Goooler: `enableJetifier ` collocate with `useAndroidX ` to enable AndroiX

gradle.properties
Goooler: `android.injected.testOnly=false ` can enable install debug apk instantly

gradle.properties
Goooler: this could be off, I turn it on because unit tests failed just now, I tune it off and try again

gradle.properties
JakeWharton: Right but we don't need jetifier because we don't have any legacy support library dependencies so it should be set to false.

gradle.properties
Goooler: image
@JakeWharton  It might something depends on AndroidX, should keep
```
android.enableJetifier=true
android.useAndroidX=true
```

Goooler:
@JakeWharton Done.

gradle.properties
```diff
@@ -1,3 +1,8 @@
+android.enableJetifier=false
+android.useAndroidX=true
+org.gradle.jvmargs=-Xmx2048m
+org.gradle.parallel=true
```
JakeWharton:
```suggestion
```

PR #588, opened 2022-10-29
Use Objects.requireNonNull().

Isira-Seneviratne:
`Objects.requireNonNull()` is desugared for API levels < 19.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.2.2'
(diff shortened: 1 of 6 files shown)


hsy719007146:
这是来自QQ邮箱的假期自动回复邮件。您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

JakeWharton:
Thanks for the PR, but this project is mostly finished in the sense that it doesn't have ongoing development. Normally I would accept such a change (minus the build changes), but we're not going to release a new version for this. Ideally this project will not need any more releases in the future and can just happily sit as a completed thing.

